PR #6, opened 2024-12-20
[Bugfix] Don't hide Google sign-in on WKWebview. 

katherineqian:
## What is in this PR?

Don't hide Google sign-in on `WKWebview` -- instead, show a button that launches app.frankly.org in the true browser.

## Changes in the codebase

* Adds a separate button to be shown on `WKWebview`s. 

* None.

## Testing this PR

Use staging link to observe that Google sign-in should work:
* On web (as normal)
* On iOS Safari
* When opened in a WKWebview (e.g. from Slack), a different button saying "Open browser for Google {signup/login}" should appear and when pressed, take you to app.frankly.org on web.


jencompgeek (approved):
Code looks good. Staging link expired so I can't really test it out, but the approach sounds good to me.

PR #7, opened 2025-01-09
Reorganize client codebase using feature-first architecture

jencompgeek:
## What is in this PR?
This is the last of the PRs for reorganizing code to create a clearer file structure. I recommend reviewing the individual commits other than the 'Restructure Client Codebase...' one to see substantive changes (of which there aren't many). The Restructure commit is nothing but file moves. For that, I would just look at the new structure of the code in browser/code editor.

## Changes in the codebase
Pretty much every file moved, as well as two monolithic classes broken down into multiple feature-specific files: `community/utils.dart` and `cloud_functions_service.dart`

None

## Testing this PR
Manual tested some basic functionality. Should be able to poke around and ensure things remain the same using the preview URL.

## Additional information
There are still some circular dependencies here I don't love. Will open a follow-up task to get rid of services.dart, which is a giant one. Also, most of the features depend on the "community" feature and the "user" feature, and there are some files in those folders that depend on other features (creating sort of a broad circular dependency). Open to suggestions on how to eliminate that (of it is even a priority to do so). Ultimately, I placed the files where they made sense: by feature/domain or in core for truly generic widgets and services and common UI elements like navbar.


katherineqian (approved):
This is such great work, Jen! The way you've organized everything makes a lot of intuitive sense to me, and I always appreciate the clarity of your work. Thank you! Just the few comments, but otherwise greatly looking forward to using this :)

PR #15, opened 2025-01-15
Add community function tests

jencompgeek:
## What is in this PR?
New tests of community-related Firebase functions

## Changes in the codebase
- Separated auth utils from firestore utils to support testing. Allows for mocking of auth functionality while still using firestore.
- Added 11 new tests of remaining community functions

None 

## Testing this PR
Tests run in CI


katherineqian (approved):
LGTM! My only question is from one that got left behind on the migration repo, which was whether some `Event`/`Community`/`User` etc. objects could be extracted out into some reusable constants for better readability/maintenance. Otherwise looks great!

PR #38, opened 2025-02-05
Fix participant removal for all event types

### LINKED ISSUE #27: Participant removal for all event types
# Details

Currently, hosts can remove participants from some but not all event types. Need to determine which event types those are and build the functionality into the others.

# Acceptance Criteria

- [ ]  User with Facilitator, Moderator, Admin or Owner privileges can remove a participant from any event type.
- [ ]  Removed participant cannot re-enter event.

jencompgeek:
## What is in this PR?
This PR fixes the following issues to ensure facilitator+ roles can kick out any participant other than the event creator from all event types:

1. A facilitator has the admin panel and can attempt to kick out a participant, but only moderator+ has permissions in the cloud function, so they receive a message that they aren't authorized to kick. Can modify the function to allow facilitator+ instead.
2. You cannot successfully set the creator's status to banned in the DB (or remove them from the meeting), but the function to kick them from agora still executes. It should not.
3. The call to revoke join_channel perms in Agora is failing every time because the parameter name "appid" is case-sensitive and is being passed in as "appId."

## Changes in the codebase
See above

None

## Testing this PR

- Kick participants from event as the facilitator. Verify immediately kicked (sees completion screen), sees a message that they are banned on event page, and is not able to rejoin.
- Attempt to kick event creator and verify creator is not kicked and can successfully participate in and rejoin meeting.

### DIFF
--- a/client/lib/features/events/features/live_meeting/features/admin_panel/presentation/widgets/admin_panel.dart
+++ b/client/lib/features/events/features/live_meeting/features/admin_panel/presentation/widgets/admin_panel.dart
@@ -695,6 +695,9 @@ class __ParticipantMenuState extends State<_ParticipantMenu> {
           if (!(confirmResult?.kickParticipant ?? false)) return;
 
           final event = EventProvider.read(context).event;
+          if (event.creatorId == widget.kickedUserId) {
+            throw VisibleException('Can\'t kick the event creator.');
+          }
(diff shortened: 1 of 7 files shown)


katherineqian (approved):
LGTM! Tried kicking participants in a test event on preview and it worked great 👍

PR #109, opened 2025-03-11
(Re-opened) Remove UI Migration widget.

### LINKED ISSUE #81: Remove UIMigration widget.
This wrapper adds custom styling on top of the app theme (on a few select pages/widgets only), which we need to remove in order to successfully apply the Material Theme to the entire app. 

### Definition of Done: 

- [x] The `UIMigration` widget is removed from the codebase.
- [x] All relevant elements applied by `UiMigration` are applied to the top-level app theme itself.

katherineqian:
## What is in this PR?

This PR was re-opened due to a Git mistake. 

Tech debt item that addresses #81 -- see description there. 

This PR is necessary because the `UIMigration` currently applies additional theming on top of the app theme, which leads to fragmentation in where the theme is defined. The goal of the overall effort is to have a single source of truth that controls the entire app theme in one place. 

**Note:** Review by commit is recommended for this PR.

## Changes in the codebase

* Created a new file called `theme.dart` to house the definition of the top-level theme.
* Applied the Text theme and Icon theme that were defined in `UIMigration` to `theme.dart`.
* Removed 65 instances of `UIMigration` across the code base, including deleting the widget itself, removing imports, and unwrapping widgets that had it. The `UIMigration` widget formerly wrapped the entire app.

* None

## Testing this PR

Open staging link and observe that nothing has been significantly altered by the removal of `UIMigration`. Visual changes resulting from this, if any, should be minor, even though the code change is significant (as `UIMigration` was a top-level wrapper of the app, and many of its subcomponents).

### Example of change
<div class="flex">
Screenshot 2025-02-27 at 14 59 26
Screenshot 2025-02-27 at 14 58 59
</div>

These before and after screenshots show an example of the change that happened when removing `UIMigration`. In both cases, the top level app theme had already declared the primary color should be black. This PR helps to allow those changes to propagate across all widgets.

### Re: Light and dark mode
* `UIMigration`'s `whiteBackground` parameter was applied rather unevenly in some cases, so the removal of it also has the benefit of helping to standardize whether widgets are themed with `Brightness.light` or `Brightness.dark` -- see the `whiteBackground` parameter of `UIMigration` for more detail. See key example below of previous `whiteBackground` parameter inconsistency:
Screenshot 2025-02-27 at 12 24 50


epenn (approved):
Ahh, gotcha. I saw that but read the name of that action a bit too literally. 😅 Good to know, thanks!

I logged and visually confirmed so this should be good to go. Sorry again for the delay on this!

PR #137, opened 2025-03-27
Merge "join event" buttons and resolve behavior inconsistencies.

### LINKED ISSUE #112: Pre-event CTA Not Working for Admins

1. Create event with a pre-event CTA
2. Enter event

### Expected behavior

Pre-event CTA should pop up after after clicking Enter event

### Actual behavior

No Pre-event CTA appears

### Additional details

Tested as an admin on Brave and as a Participant on Chrome

Brave & Chrome

### Operating system

MacOS 14.4.1

katherineqian:
### Behavior unification
The two Join Event buttons had two distinct code paths, leading to inconsistent and buggy behavior on top of being difficult to read and maintain. This PR resolves those inconsistencies: 
* On current staging, try setting an event's platform to an external provider (e.g. Google Maps, Zoom). Click the purple and green join event buttons and observe that the green button still takes you to Frankly despite having set a different provider, while the purple button correctly takes you to the external provider.
* Analytics events were being inconsistently logged: 
    * Entering the event via the green button did not generate an "Enter Event" analytics event, so only users who entered via the purple button would be counted. 
    * RSVPing for an event was not being logged for event creators. 
* CTA behavior inconsistency: The purple join event button would only show CTAs once (and not ever for event creators, unless they created from a template), while the green join event button would always show CTAs no matter what. 

### Bug fix and de-duplication
Fixes #112 by standardizing when CTAs should be shown. Even within the same button, the logic for what would happen when "join event" was pressed was very complex. Duplicated logic has been cleaned up / clarified.

### RSVPing
In order to fix the behavior shown to event creators, we've decided that event creators should also RSVP to their own events so that they go through the same flow as end users. 

As a result, the RSVP call has been removed from the create/update event flow. 

Also, RSVPs are only shown if it's >= 10 minutes before the event. The RSVP dialog is not shown after the event started or ended (currently it's still possible to join an event after it has ended, with no time limit, but the RSVP is not the place to address this problem).

## Changes in the codebase

* Removed second code pathway from the "green" Join Event button (which was defined in the method `_buildEnterEvent`). These buttons will now both call the same function and should have the same behavior.
* Removed duplicated code from `_joinEvent` and `_startMeeting` methods, which also led to inconsistent behavior. Now that code is all contained within the `_joinEvent` method.
* Moved analytics events to their logical homes so that they will be consistently logged whenever actions take place (RSVPing to an event is logged in `verifyAvailableForEvent` and entering an event is logged in `_joinEvent`). 
* Fixed bug where external provider is inconsistently used. 
* Introduce new variable in `_joinEvent` which checks if this is the user's first time joining an event.
* Standardize when RSVPs are shown:
    * Remove the RSVP check during event creation flow
    * Only show RSVP if it's >= 10 minutes before the event. 
    * **Note:** RSVP is also not shown after the event. 

* Removes behavior where event creator is automatically added as an active participant during the event creation call. Now event creators will go through the same UX flow as users where they must RSVP to their own event and see the CTA after RSVPing. 

* None. 

## Testing this PR

This PR should resolve previous inconsistent behavior. The following scenarios should be tested on both the "purple" (left-hand side) and "green" (appears on top of the meeting preview image and says "Enter Event") Join Event buttons:
- [x] Create an event, then:
    - [x] Add a CTA to the event. 
    - [x] Try to join the event, but select "no" when RSVPing. Observe that you should not be redirected to the meeting.
    - [x] RSVP to the event and observe that the CTA should be shown. 
    - [x] Join the event a second time and observe that no CTA should be shown. 
- [x] On another account, RSVP/join the same event and observe that a CTA should be shown.
    - [x] Join the event a second time and observe that no CTA should be shown. 
- [x] Choose a different meeting provider (in event settings, click "Frankly video" and choose a different provider like Google Maps). Click both buttons and observe that the external provider should be launched instead of Frankly's built-in meeting.
- [x] RSVP should only be shown when it's not within 15 minutes of the start time (i.e. should not be shown if the event is starting in >15 minutes). 
    - [x] Test that RSVP is not shown when within 15 minutes of the start time
    - [x] Test that Participant still "joins" event if they directly join within 10 mins without an RSVP 

## Additional information

**Known issue:** After user RSVPs, there is some loading time before the pre-event CTA is shown. If the user navigates away during this loading segment, the pre-event CTA will never re-appear.


saihla:
I tested this behavior and the Pre-survey is appearing exactly once for both admins and participants as intended. However, it's also sending me _into the event_ immediately upon registration (both as an admin and as a participant) even when the event hasn't started yet. We don't want people to enter the event until the "enter event" option shows up.

saihla:
@katherineqian I tested this behavior on Brave (as an admin) and Firefox (as a participant) and it all looks to be working well! The only issues I ran into were out of scope and already added to GH (e.g. https://github.com/berkmancenter/frankly/issues/220, [#221](https://github.com/berkmancenter/frankly/issues/221). Thank you so much for cleaning this up!

saihla (commented):
Reviewed here in Aug: https://github.com/berkmancenter/frankly/pull/137#issuecomment-3166133841

Approved from product perspective. Not approving in GH because it still needs code review.

epenn (approved):
- Ran through the testing checklist above and everything looks good
- Appreciate some of the additional error handling as well
- And good call re: event creator
- LGTM!

PR #147, opened 2025-04-01
Update template page to follow theme.

katherineqian:
## What is in this PR?

The purpose of this PR is to remove all instances of hardcoded `AppColor` and replace them with their equivalents in the new Material Theme. In other words, all views in the Template page should now pull from an app Theme instead of being hardcoded into specific widgets.

## Changes in the codebase

* Removed and replaced all hardcoded `AppColor` and `Color` instances with reasonable equivalents from Material theme
* Refactored several widgets to be simpler (and therefore more maintainable/performant)
* Removed several instances of isWhiteBackground parameter which was unused throughout the codebase

## Changes outside the codebase

* N/A

## Testing this PR

Open staging link and create a template. The following pages should be updated and follow theme colors:

**Template list page**: follows community-defined theme
Screenshot 2025-04-01 at 15 25 17

Template home
Screenshot 2025-04-01 at 15 37 47

Editing agenda items
Screenshot 2025-04-01 at 15 40 59

Editing CTAs
Screenshot 2025-04-01 at 15 41 33

Editing Template Sidebar
Screenshot 2025-04-01 at 15 43 32

Editing Template Settings Sidebar
Screenshot 2025-04-01 at 15 43 45

## Additional information

N/A

PR #151, opened 2025-04-08
Remove AppColor.darkBlue.

### LINKED ISSUE #163: First pass at removing hard-coded colors
This task involves a first pass at swapping hard-coded color values with appropriate colors from the app theme.

katherineqian:
## What is in this PR?

Instead of refactoring page by page as in #147, the goal of this PR is to simply replace `AppColor.darkBlue` as part of a "first pass" strategy for the theme migration. The goal of using this strategy is to make it easier to start parallelized work on theme refactoring.

## Changes in the codebase

* Replaced all instances of `AppColor.darkBlue` with `context.theme.colorScheme.primaryColor`.
* Refactored code where accessing context was previously impossible - now all former `darkBlue` instances can pull from context. **Note:** compared to prior refactoring, most of the existing code was kept intact here, only doing as little refactoring as was needed to properly access context.
* Updated all imports to be able to pull from context theme.
* Removed duplicate imports and fixed some lint errors.

* None.

## Testing this PR

Not much should be observed visually other than that the lime-green version of `AppColor.darkBlue` should now no longer be visible anywhere in the app. `AppColor.darkBlue` should also no longer show up in the codebase anywhere.

## Additional information

To come: 
* Removing all the other `AppColor`s and `Color`s get to a state where that class is no longer used (or at least in only very minimal, rare, occasions)
* Implementing the updated M3 color theme and text theme from Figma


epenn (approved):
Confirmed `AppColor.darkBlue` is no longer present in the code base and on a click through of the preview site lime green couldn't be found. LGTM!

PR #155, opened 2025-04-11
Define Material 3 Theme.

### LINKED ISSUE #120: Implement Material 3 Text Theme.
We are using a new font family! This task involves updating the theme text styles.

katherineqian:
## What is in this PR?

The purpose of this PR is to define a top level Material 3 Theme that will allow us to use M3 widgets in Flutter with out-of-the-box styling that follows our Figma files.

## Changes in the codebase

* Use the Figma exported M3 theme to define a color scheme (light for now).
* Based on the Figma file, define a Text Theme for the app.
* Use the new themes.

* None.

## Testing this PR

The staging link shows very minimal change based on this PR because many styles are manually defined in this app (color and themes). However, if you add any default M3 widget, you should observe that the widget automagically gets styled to look exactly like our Figma file!

### DIFF
--- a/client/lib/styles/material_theme.dart
+++ b/client/lib/styles/material_theme.dart
@@ -1,0 +1,572 @@
+import 'package:flutter/material.dart';
+
+class MaterialTheme {
+  final TextTheme textTheme;
+
+  const MaterialTheme(this.textTheme);
+
+  static MaterialScheme lightScheme() {
+    return const MaterialScheme(
+      brightness: Brightness.light,
+      primary: Color(0xff000000),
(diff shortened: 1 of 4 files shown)


epenn (approved):
Looks good! Clicked around for a bit and observed that the app generally appears/works as expected after applying the Material theme update. And I'm happy to take your word that the hundreds of hex color values in material_theme.dart match what was in Figma. 😀

Also re: your note about adding M3 widgets, it might be worth adding the screenshots you took previously of the OutlinedButton showing that it renders what was in Figma as a comment here since it demonstrates what you werer pointing out.

PR #167, opened 2025-04-23
Refactor `ActionButton` to use Material 3 buttons and theme colors.

katherineqian:
## What is in this PR?

This PR aims to remove as much additional **style** code on the `ActionButton` as possible so that the button can fall back to Material defaults, which is especially helpful for handling different state styles (which comes out of the box). It also contains some quality-of-life naming updates. 

## Changes in the codebase

* Importantly, removed the `ButtonStyle` method of styling the action button and replaced with `.styleFrom`, which allows us to let Material handle the styling of different button states for us automatically.
* Removed several default colors which were calculated by the button, in favor of using the default theme style unless a property was specifically manually provided as a parameter (e.g. `textColor` is now used to style the `foregroundColor` on buttons, but falls back to the default style in our M3 design system if not provided).
* Removed `overlayColor` parameter
* Implemented a new `ActionButtonType`, the `TextButton` which is part of the M3 system and will likely be used going forward.

* None.

## Testing this PR

Hard to test other than just visual, plus some fixes are still ongoing -- however, on staging, buttons should generally appear as primary buttons with different styling for different states.

### DIFF
--- a/client/lib/core/widgets/buttons/action_button.dart
+++ b/client/lib/core/widgets/buttons/action_button.dart
@@ -96,11 +96,7 @@ class ActionButton extends StatefulWidget {
     this.height,
     this.tooltipText,
     this.child,
-  })  : assert(
-          textStyle == null || textColor == null,
-          'Cannot specify textStyle and textColor',
-        ),
-        assert(child == null || text == null, 'Cannot specify child and text'),
+  })  : assert(child == null || text == null, 'Cannot specify child and text'),
         super(key: key);
 
   @override


client/lib/core/widgets/buttons/action_button.dart
```diff
@@ -158,42 +149,23 @@ class _ActionButtonState extends State<ActionButton> {
...
-          color: _getTextColor(),
+          color: widget.textColor,
         ),
       );
     }
 
     throw Exception('Icon must be a Widget or IconData instance.');
   }
 
-  Color _getTextColor() {
```
johnnycrich, line 169: LOVE seeing this go away.

client/lib/core/widgets/buttons/action_button.dart
```diff
@@ -230,56 +200,61 @@ class _ActionButtonState extends State<ActionButton> {
...
-        ),
-        child: _buildButtonContents(),
-      );
-    } else if (widget.type == ActionButtonType.outline) {
-      button = OutlinedButton(
-        style: ButtonStyle(
-          overlayColor: overlayColor,
-          side: WidgetStateProperty.all(
-            widget.borderSide ??
+    switch (widget.type) {
```
johnnycrich, line 213: Refactoring as a `switch` such a good move!

johnnycrich (approved):
This looks great on my end!

PR #174, opened 2025-05-03
[Checkpoint] Theme update, sign-in flow, and internationalization updates.

katherineqian:
## What is in this PR?

This PR aims to bring the app to a usable and testable intermediate state, removing any color swaps that led to unusable/unreadable text, icons, buttons, etc. in the app. 

## Changes in the codebase

* AppTextStyle class was swapped to Inter. 
* Color changes in several parts of the app.

* None.

## Testing this PR

Use the app and hopefully observe that all text and buttons should be readable (not invisible or extremely low-contrast).


epenn (approved):
Approving so we can merge into the staging branch and collectively test for regressions with previous updates already queued up

PR #181, opened 2025-05-16
Theme color updates.

### LINKED ISSUE #180: Update primary and surface colors.
The primary color should be updated to a slightly lighter color so that hover effects are visible on it. The surface color should also be updated to a more neutral (less red) variation.

katherineqian:
## What is in this PR?

Updates surface color to be neutral and primary color to be slightly lighter so that hover effects are more visible to the eye. 

## Changes in the codebase

* Updates primary and surface colors in Material theme file `material_theme.dart`. 

* None.

## Testing this PR

Test the staging link and observe that appropriate colors should be updated.
**Note:** Button on sign-in flow appears to not have the updated hover effect, so this should be updated at some point to follow themed buttons.


epenn (approved):
LGTM!

PR #188, opened 2025-05-30
[Bugfix] Fix text fields losing focus immediately after tapping on them.

### LINKED ISSUE #189: Text fields require multiple taps to focus

1. Try tapping on any text field in the app (e.g. sign in dialog), not directly in the center of the field.
2. Observe that the cursor immediately disappears when you release the mouse button.
3. Observe that the text field requires multiple taps to focus.

See video below.
https://github.com/user-attachments/assets/26f00983-4c8c-46e5-be02-57483362d101

### Expected behavior

I should be able to tap on a text field just once to start typing. 

### Actual behavior

I need to tap multiple times.

### Additional details

_No response_

whatsmybrowser.org/b/39L8G

### Operating system

_No response_

katherineqian:
## What is in this PR?

In the current platform, in most/all cases, TextFields require _multiple_ taps in order to get them focused (in order to start typing). Upon closer inspection, this happens if the user taps anywhere within the text box that is not also within the height boundary of the cursor.

This PR fixes that bug by removing the `unfocusOnTap` method within the `FocusFixer` wrapper that was previously applied to the entire app, which caused an immediate unfocus on tap behavior that led to this buggy effect.

## Changes in the codebase

* Removed `unfocusOnTap` parameter and associated code.

* None.

## Testing this PR

Based on the original comment:

> /// Listens for pointer up events and calls to remove the focus. This dismisses the keyboard and focus
> /// if the user taps outside of the keyboard or text field

This code appears to be made for dismissing the mobile keyboard when tapping outside a TextField. After removing this code, it looks like the keyboard is still properly dismissed when tapping outside -- reviewer can verify. No regressions have been observed on the staging link for mobile web so far.


epenn (approved):
This looks good to me! I confirmed focus is no longer lost when tapping/clicking on text fields (regardless of where on the text field you tap). Works as expected on desktop and mobile using Chrome, Safari, and Firefox on both. No obvious regressions when clicking around through text fields across the platform. And mobile keyboard disappears as expected as well.

I did find one (small) oddity on mobile Safari that I don't *think* is related, but testing this is what made me notice it. On the login page, when you focus on either of the email or password textboxes, the mobile keyboard pops up and the bottom bar with the "Log in" and "Sign Up" moves to the top of the mobile keyboard just as you would expect. But when you remove focus and the keyboard disappears, the bottom bar hovers in the middle of the page for roughly a full second while (seemingly) the keyboard animation finishes up. This seems to be limited to mobile Safari. Chrome and Firefox move the bar instantaneously as expected.

PR #222, opened 2025-08-05
Decouple A/V device settings from Agora.

### LINKED ISSUE #59: Super close-up on iOS
Tested on Chrome and Safari on iOS. In both cases the camera was suuuuper zoomed in, even when phone was held at arm's length.

[iPhone SE 2020 running iOS 17.6.1. Yes my phone is old.]

Please enjoy the silly selfies. Screen shots taken on desktop so you can see how far away I was holding my phone.

Image
Image

### LINKED ISSUE #176: Decouple A/V Handling from Agora
**Summary:**
There are a few issues with the way Frankly manages camera and microphone settings. Rather than fixing them one by one, we're considering a rebuild of this part of the app so that Frankly can manage things like permissions, muting, and device selection on its own — instead of relying on a third-party video service. This could make things more reliable and flexible, and would also make it easier to support different video platforms in the future.

---

# Overview:
There are ongoing performance issues and minor bugs related to audio & video device selection, browser permissions handling, and responsiveness within A/V settings.

Instead of addressing each of these issues individually, this may be a good opportunity to refactor the implementation of media device management within Frankly.

By decoupling camera and microphone selection, permission handling, mute controls, and related functionality from Agora, we can create a service-agnostic media layer. This would improve interoperability with other video conferencing solutions (e.g., Zoom SDK, Jitsi), making it easier to integrate alternative services or switch providers as needed.

**What we are considering handling natively within Frankly:**

- Permissions – Request access to camera and microphone (i.e. handle device/browser permissions)
- Device enumeration – List available microphones and cameras
- Device selection – Allow participants to choose their preferred mic and camera
- Device toggling – Support muting/unmuting and turning camera on/off

---

# Considerations
Before committing to this approach, we need to determine how feasible this is and weigh the pros & cons.

### Pros:

- **Opportunity to Replace Legacy Implementation**
Gives us a clean slate by avoiding the need to patch existing tech debt or address known bugs and performance issues one by one. A fresh, centralized approach can lead to a more maintainable and scalable solution long-term.

- **Increased Flexibility & Reduced Vendor Lock-In**
By decoupling device and permission handling from Agora, we gain more control and make it easier to switch providers or support multiple conferencing platforms without being tightly coupled to any one of them.

- **Customizable UI/UX**  
Gives us full control over how device-related interactions look and behave (e.g., toggle buttons, device dropdowns, permission messaging).

### Cons:

- **Upfront Engineering Effort**  
This would require custom work to replicate functionality that Agora currently handles for us.

- **Ongoing Maintenance Responsibility**  
We’d be responsible for keeping up with browser/OS changes around device and permission APIs.

- **Possible Trade-Offs in SDK-Level Features**  
We may lose access to features like echo cancellation or noise suppression if they’re only applied when using the SDK’s capture process. This would be a strong consideration, especially if we ever want to use Zoom's SDK.

### LINKED ISSUE #215: (Intermittent) Active device does not match selected device

Sometimes the camera that appears selected in the A/V settings is not the camera actually being used. You need to toggle to a different camera and then back to the one you want for the setting to be accurate.

Specifically in my case, Frankly loves to default to using my iphone's camera. (This is not something I experience in other video platforms.) Frequently it will use my phone's camera even though the A/V settings show my webcam selected. To resolve it I need to change the setting to the iphone camera (which in reality is already active) and then toggle back to my webcam to get it to switch.

The fact that it defaults to my iphone camera is strange but not necessarily a bug. The fact that the selected device doesn't match the active device is the bug.

Failed attempt to repro: https://www.loom.com/share/2d44f62fb7994314a945d47b6a410d09?sid=e8a8f761-2e32-40ba-8f89-5e13e2176efe

Successful attempt to repro: https://www.loom.com/share/ad365f0fb5c64cecb2aed19e22c0f60a?sid=8cd57922-6d53-4278-8fed-6998bc4fb1ef

I'm including the failed repro attempt in case the fact that I adjusted my settings in the first attempt impacted the settings in the second attempt.

### Expected behavior

Selected device in A/V settings should reflect the device being used.

### Actual behavior

Selected device in A/V settings is sometimes different from the device being used.

### Additional details

I've only observed this with the selected camera, I don't know if this happens with microphones as well. That might be harder to repro because the system already defaults to the microphone I want.

Brave / Chromium. Not including version since this has been happening for a while.

### Operating system

MacOS 14.4.1 (23E224)

### LINKED ISSUE #227: A/V options show up blank

1. Join a meeting
2. Open the Audio/Video settings options menu. 
3. Observe that the dropdown menus for both audio and video appear blank as shown in the screenshot below.

Image

Note: This issue appears to be intermittent, i.e. can't be reproduced on every single run of the app. 

Nit: There is also a bright blue border hard-coded into this dialog. 

### Expected behavior

I should be able to see audio/video device options in this dialog. If empty, it should display that the options are empty.

### Actual behavior

Dropdown menu is blank.

### Additional details

_No response_

Chrome

### Operating system

macOS

### LINKED ISSUE #228: Video flashing on iOS

1. Join a meeting on iOS device.
2. Observe that video flashes black every few seconds. See video below.

https://github.com/user-attachments/assets/a2b89c3c-ac81-4c6c-bc76-49a4bcdd1f34

### Expected behavior

Video should display normally.

### Actual behavior

Video is flashing rather disruptively.

### Additional details

_No response_

Firefox, Safari

### Operating system

iOS

katherineqian:
## What is in this PR?

This PR closes #176. The goal is to implement an "in-house" service that manages device selection and permissions independently of the video service provider (in this case, Agora). This sets the foundation for making the platform more flexible and extensible, as previously all interactions with A/V were mediated through Agora function calls.

Agora for Flutter Web is currently in alpha mode. As a result, there were several roadblocks (especially in mobile web browsers) in the implementation of this PR, which necessitated some engineering workarounds. For example, the `setDevice` method for audio and video are noted as "for Windows and macOS only" in the documentation. The `setDevice` call for video did not reliably work until the video was toggled off and on again.

Additionally, the main roadblock that prevented us from fully decoupling media handling from Agora was that the `.pushVideoFrame` method is not supported on Flutter Web -- see issue filed on the SDK [here](https://github.com/AgoraIO-Extensions/Agora-Flutter-SDK/issues/2389) -- the method not being supported was confirmed with an Agora team member.

For desktop Web, a video preview functionality has been introduced that is independent of Agora (uses only the `MediaDeviceService`. Unfortunately, this had to be disabled for mobile due to the fact that `getUserMedia` can only support one camera at a time on iOS ([Bugzilla documentation](https://bugs.webkit.org/show_bug.cgi?id=238492)). `getUserMedia` is already used in the underlying Agora video implementation which pushes video to the remote service (which we cannot decouple from for reasons stated above). 

**This PR builds off the initial foundational work done by @bestian in #193. Thank you!** 
Note for @bestian: I was originally going to merge this branch into your branch, but it seems like that may cause some additional overhead, since I branched off of an earlier state in your branch. If it's okay with you, I think it would be easiest to merge this branch into staging when ready -- your commits are still attributed to you in this branch. Please let me know if you have any other suggestions!

## Changes in the codebase

* Creates a new `MediaDeviceService` that manages, A/V devices, permissions, and introduces a new video preview functionality (only on desktop) that is independent of Agora. 
* Creates a new `MediaSettingsWidget` that users can use to manage their A/V device selections, and on desktop, shows a video preview to self. Fixes the styling issues on the previous A/V settings widget.
* Fixes a race condition in `agora_room.dart` that led to video device intermittently being wrong when joining a meeting -- should fix #215. See changes made to the `RtcEngineEventHandler` in `agora_room.dart`. 
* Fixes a race condition that led to permissions sometimes not being requested before joining a meeting. Should close #227. 

* None.

## Testing this PR

Please see the internal [A/V QA](https://www.notion.so/rebootingsocialmedia/A-V-Q-A-247d125414f780a48b0ece20ad935f9b#247d125414f780a48b0ece20ad935f9b) tracker for all behaviors that should be tested for this PR.

**Note:** The staging link is somewhat hidden amidst the commits that were made for this PR. [Here](https://gen-hls-bkc-7627--pr222-kq-feat-decouple-av-utlxr0h2.web.app/) it is for convenience.

## Additional information

**Known issues**
* Selecting video devices other than the default FaceTime camera does not work on Firefox ESR for macOS. This is an existing issue on prod and not introduced by the changes in this PR. 
* On Safari, the video still may have an issue with flickering intermittently. This is difficult to reproduce, but it is also an existing bug.


bestian:
> Note for @bestian: I was originally going to merge this branch into your branch, but it seems like that may cause some additional overhead, since I branched off of an earlier state in your branch. If it's okay with you, I think it would be easiest to merge this branch into staging when ready -- your commits are still attributed to you in this branch. Please let me know if you have any other suggestions!

hi @katherineqian,

thank you for solving this huge task.

It' fine for me to merge this branch into staging directly when ready.  

my branch `decouple-a-v-handling` with overgrown codes can be shelved.

client/lib/core/utils/media_device_service.dart
```diff
@@ -0,0 +1,188 @@
...
+
+  Future<void> init() async {
+    try {
+      // Start by requesting permissions so that devices can be listed.
+      micPermissionStatus = await requestPermissions(Permission.microphone);
+      cameraPermissionStatus = await requestPermissions(Permission.camera);
+
+      final devices =
+          await html.window.navigator.mediaDevices?.enumerateDevices();
+      if (devices != null) {
```
dariusk: So I think that if there are no available devices, the return value from `enumerateDevices()` will be an empty `Array` rather than `null`. [MDN says the resolved return value is always an array of `MediaDeviceInfo` objects](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices#return_value), which implies to me it would be an empty array if there were no devices. And then if we look at [the specification for `enumerateDevices`](https://w3c.github.io/mediacapture-main/#dom-mediadevices-enumeratedevices), step 12 is always to return a _resultList_, which in our context is an `Array`.

So I guess my suggestion is:

```suggestion
      if (devices?.length > 0) {
```

with the optional chaining operator to catch cases where this assumption is wrong or misimplemented.

I would test this myself but I am not sure how to emulate a device with no camera or microphone!

client/lib/core/widgets/media_settings_widget.dart
```diff
@@ -0,0 +1,475 @@
...
+                      );
+                    }).toList();
+                  },
+                  onChanged: (val) async {
+                    if (val != null) {
+                      setState(() {
+                        selectedAudioDeviceId = val;
+                      });
+                    }
+                  },
```
dariusk: What happens in the case where `val` _is_ `null`? This is the error state mentioned in the comment on line 148, right? I guess the outcome here is "nothing" and I suppose that's the correct outcome. Mostly just commenting because I always wonder when there's neither an "else" nor some other code that occurs after the conditional.

I might rephrase this as

```suggestion
                  onChanged: (val) async {
                    if (val == null) return;
                    setState(() {
                      selectedAudioDeviceId = val;
                    });
                  },
```

It just makes it clearer to the reader that "on `null` we want to do nothing", and most of the time we want to `setState`.

client/lib/core/widgets/media_settings_widget.dart
```diff
@@ -0,0 +1,475 @@
...
+        children: [
+          Text(
+            'Audio Input Device',
+            style: context.theme.textTheme.titleMedium,
+          ),
+          _mediaService.audioInputs.isEmpty
+              ? Padding(
+                  padding: const EdgeInsets.only(top: 16.0),
+                  child: Text(
+                    'No audio devices available. Please check permissions.',
```
dariusk: I won't comment on each of these, but all the strings here and elsewhere should be pulled from [our l10n definitions](https://github.com/berkmancenter/frankly/tree/staging/client/lib/l10n), yeah?

client/lib/core/widgets/media_settings_widget.dart
```diff
@@ -0,0 +1,475 @@
...
+                      });
+                      await _mediaService.selectVideoDevice(
+                        deviceId: val,
+                        shouldUpdatePreview: widget.shouldShowVideoPreview,
+                      );
+                      await updatePreview();
+                      setState(() {
+                        isLoading = false;
+                      });
+                    }
```
dariusk: Same as my comment above, I think doing `if (val == null) return;` and then following it up with the rest of the code to be executed otherwise is more readable.

client/lib/features/events/features/live_meeting/features/video/presentation/widgets/participant_widget.dart
dariusk: Wow, I love the overall refactor in this file, what an improvement!!!

dariusk (commented):
Wonderful work overall! FWIW this was a "naive" review -- I don't really know much about Agora though I do know about browser APIs, so I did my best to comment where things didn't make sense or lit a warning light for me as an outside reader.

client/lib/core/widgets/media_settings_widget.dart
```diff
@@ -0,0 +1,475 @@
...
+      (int viewId) => _videoElement,
+    );
+  }
+
+  Future<void> initAll() async {
+    await _mediaService.init();
+
+    initialAudioDeviceId = _mediaService.selectedAudioInputId;
+    selectedAudioDeviceId = _mediaService.selectedAudioInputId;
+    initialVideoDeviceId = _mediaService.selectedVideoInputId;
```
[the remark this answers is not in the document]
katherineqian: This is somewhat of a relic of a past implementation. For video preview, when the user selects a device, we need to `getUserMedia` to show the updated video. But for microphone, since there is no preview, we don't really have to update the selected mic in `MediaDeviceService` until the user hits "save", so I had used this variable to hold the user's selection instead.

However, I agree it's a little confusing/not as maintainable for it to be inconsistent like this. I updated the implementations to both interact directly with `MediaDeviceService`.

dariusk (approved):
Awesome, this all looks good to me -- most of the tricky fixes are all boolean logic and order-of-operations with initialization, and I know we've run those through their paces in staging and staging does in fact work, so I'm happy with what we have here! Plus the strings are localized, and I really like how `ParticipantGridLayout` simplifies its parent components.

Approved!

saihla (approved):
I'm so excited for this to go live!

For anyone external seeing this - As part of QA testing for this substantial work, we've help many team syncs on this branch. The A/V during these events have been so much smoother!

Issue #202, opened 2025-06-23
Text Field Improvements


saihla:
**Loom:** [Improving Text Field Behavior](https://www.loom.com/share/5b35daa3e61c4e15930acdd20ee65fb2?sid=5b7fc170-e372-4749-a1cf-1347ac521346)

**1. Ensure elements (including surrounding elements) don't move on focus.**
On focus, the text field border gets thicker and in some cases the character counter appears (correct behavior). I'd like to ensure that when this happens, the overall space that the element is taking up doesn't change, so that elements aren't jostled around based on state.

**2. Expand multiline fields to fit content**
When multiline content is longer than the height of the text field, expand the field instead of adding a scroll bar.

_Note: I expect this is preferred behavior for all multiline fields, but if there are cases where this doesn't make sense please flag them. The main example I'm thinking of is agenda item content, where it's frustrating to not be able to see all of it._

PR #241, opened 2025-10-10
Text field improvements

johnnycrich:
## What is in this PR?

As per changes requested in #202, allows expansion of multiline fields to fit content, and resolves the issue where UI rendering would slightly shift if text counter is visible under text field.

## Changes in the codebase

* On `CustomTextField`, `null` as an allowable value to `maxLines`, causing field to expand indefinitely as line breaks are added. A number here will allow expansion up to that limit.
* Check focus state in text counter rendering; if focus is not on field, an empty box is rendered at equal height. 
* Removed superfluous Container from community text fields.
* Applied maxLines to multiline text fields where relevant.

## Testing this PR

- [x] Edit a community and note that the About field expands as text is entered. There is no theoretical limit to this expansion (yet?).
- [x] The same behavior should now happen for agenda item or event CTA body fields.
- [x] Try writing an announcement to an event and note that the pop-up text field has a limit of 15 lines, at which point it stops expanding and reverts to scrolling behavior
- [x] Note that all text fields with a counter no longer propagate a small layout shift between focus in/out states.

### DIFF
--- a/client/lib/core/widgets/custom_text_field.dart
+++ b/client/lib/core/widgets/custom_text_field.dart
@@ -11,10 +11,10 @@ enum BorderType {
   outline,
 }
 
-/// A customizable text field widget. 
+/// A customizable text field widget.
 /// This widget provides various options to customize its appearance and behavior,
 /// including border type, padding, text styles, and more.
-/// 
+///
 /// {@tool snippet}
 /// ```dart
 /// CustomTextField(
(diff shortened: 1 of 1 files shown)


client/lib/core/utils/dialogs.dart
```diff
@@ -72,6 +72,7 @@ class Dialogs {
                   key: formKey,
                   child: CustomTextField(
                     minLines: 3,
+                    maxLines: 15,
```
katherineqian: Nice attention to detail on this dialog height :)

PR #242, opened 2025-10-20
Resolve camera turning on unexpectedly

johnnycrich:
This resolves the issue of a user's camera turning on when entering a breakout room, despite having disabled it previously.

## Changes in the codebase
In `ConferenceRoom`'s `connect()`, reference `liveMeetingProvider` to determine if user has previously opted to default to video to on, and pass to `AgoraRoom` `connect()`. For context, this provider is given an updated value for this whenever video is toggled.

## Testing this PR

- [ ] Create an event and enter it. If your video defaults to on, turn it off.
- [ ] As an admin, force yourself into a breakout room.
- [ ] Note that your video remains off. Now, turn it back on.
- [ ] As an admin, end the breakout session.
- [ ] Your video should still be on.
- [ ] Reset your browser A/V permissions and refresh.
- [ ] Enter the event again and try to enable your video.
- [ ] You should be given a prompt to enable video by the browser, and then your video should turn on.

### DIFF
--- a/client/lib/features/events/features/live_meeting/features/video/data/providers/conference_room.dart
+++ b/client/lib/features/events/features/live_meeting/features/video/data/providers/conference_room.dart
@@ -307,7 +307,7 @@ class ConferenceRoom with ChangeNotifier {
         eventProvider: liveMeetingProvider.eventProvider,
         conferenceRoom: this,
       );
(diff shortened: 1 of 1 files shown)


epenn (approved):
LGTM!

PR #246, opened 2025-10-27
[Bugfix] Allow default community theme to pass color check.

katherineqian:
[Bugfix] Allow default community theme to pass color check (when both colors are null).

## What is in this PR?

This fixes a regression introduced in #216 where communities could no longer be saved due to an "invalid theme" (even if the theme was not edited). 

## Changes in the codebase

* When no theme is selected, the theme colors are actually null - this PR allows that case to pass as a valid theme. 

* N/A

## Testing this PR

- [ ] Edit a new community and observe that no error message should be shown. 
- [ ] Edit the theme colors to be any of the theme presets (try all of them) and observe that no error message should be shown.
- [ ] Choose a low-contrast custom theme and observe that the theme should still be detected and shown as invalid, with the error text saying to choose a valid combination.

### DIFF
--- a/client/lib/features/community/features/create_community/presentation/views/create_community_dialog.dart
+++ b/client/lib/features/community/features/create_community/presentation/views/create_community_dialog.dart
@@ -158,11 +158,15 @@ class _CreateCommunityDialogState extends State<_CreateCommunityDialog> {
       );
     }
 
-    final isThemeValid = ThemeUtils.isColorComboValid(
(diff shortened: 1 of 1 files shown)


epenn (approved):
- LGTM!
- No error shown when creating the new community.
- No errors on any of the preset selections.
- Correct errors shown for both low contrast themes and identical color themes.

PR #251, opened 2025-11-05
Update email template colors.

### LINKED ISSUE #257: Update email template colors to match Frankly branding
Our email templates still follow a legacy color palette / branding:
Image

The colors should be changed to match Frankly branding as it is confusing to receive emails that have a completely different look:
Image

katherineqian:
## What is in this PR?

Our emails still have the legacy color palette, which is a little jarring to see especially as it is completely different from our current branding.

## Changes in the codebase

* Color swaps for all of the defined email templates.

* None.

## Testing this PR

We need to test all the different email types to ensure they are still readable after the update.

## Additional information

This email template file is far too long and extremely difficult to maintain. I've just done some color swaps for now but would prefer to refactor.

### DIFF
--- a/firebase/functions/lib/utils/email_templates.dart
+++ b/firebase/functions/lib/utils/email_templates.dart
@@ -80,11 +80,11 @@ String generateEmailEventInfo({
       <div class="section">
           <div class="center">Add to calendar:</div>
           <div class="center">
-            <a href="$calendarGoogleLink" style="color:#303B5F;"><b>Google</b></a>
+            <a href="$calendarGoogleLink" style="color:#313030;"><b>Google</b></a>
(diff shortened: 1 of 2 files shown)


epenn (approved):
I tried sending several of the email types and all appear to be readable so far. I completely agree that refactoring the templates file is a good idea as well. I'm going to mark this as approved, although @katherineqian given the number of different email variations, I'm also going to ping you when we're back next week just to check whether you think what I tested should cover this. Probably worth holding off on merging until after that.

epenn:
Just realized I didn't close the loop above, but @katherineqian and I chatted and confirmed the test emails I used while reviewing covered this.

PR #252, opened 2025-11-06
Persist auth emulator login to entire session

johnnycrich:
I banish ye, auth gremlins. 🏴‍☠️

Arr, this here be a change persistin' the current user login te an entire browser session when ye be usin' that scallywag autentin' emulator, preventin' the need to login after every 'ard refresh. Ahoy! ⚓️

## Changes in the codebase

* set persistence to `Persistence.SESSION` when using a firebase authentication emulator in `client/lib/features/user/data/services/user_service.dart`

## Testing this PR

- [ ] Apply this change and see that after logging in, you no longer have to login again after refreshing the app. 
- [ ] Close the current browser session.
- [ ] Reopen the app. Note that you have to login again.


epenn (approved):
Sorry for the delay on this one but this looks good to me!

PR #286, opened 2026-01-29
Fix poll data showing 0/0 votes in agenda overview card.

### LINKED ISSUE #245: Poll data not persisted after switching agendas — results reset to zero

1. Create or enter a meeting with multiple agenda items(a poll and a suggestion).
2. During the poll agenda, submit a vote.
3. End that agenda, then proceed to the next agenda.
4. Click back to the previous poll agenda to review the poll results.

### Expected behavior

The poll results from the previous agenda should remain visible and reflect the votes that were submitted before moving on to the next agenda.

### Actual behavior

When returning to the previous agenda, all poll options show zero results, as if no votes were ever recorded. This happens even without reloading the page — simply switching between agendas causes the poll data to be lost. It appears that poll data is not being persisted when an agenda ends.

### Additional details

Initial investigation suggests the issue occurs during the data persistence stage. Poll data is available while the agenda is active, but not stored afterward, resulting in empty results upon revisiting.

Image

Chrome 141 https://www.whatsmybrowser.org/b/XT2U2

### Operating system

macOS 15.6.1

katherineqian:
## What is in this PR?

Fixes #245 -- currently, poll data shows 0/0 in the agenda overview. This PR changes the UI displayed for poll cards in that view so that it just lists the options, without the blank poll data.

## Changes in the codebase

* UI changes to poll card in agenda overview view, based on whether poll data is passed into the widget.
* Add assertions to protect against incorrect usage of the widget.

* N/A

## Testing this PR

* Create a meeting with a poll card item in the agenda.
* Open the Agenda sidebar tab. 
* Observe that the poll card no longer shows incorrect blank data for the poll results, just the poll options. 

## Additional information

**Before (view of poll item in agenda overview)**
Screenshot 2026-01-23 at 16 59 20

**After:**
Screenshot 2026-01-23 at 17 37 27

**Note:** While this card could have been refactored to show the poll data collected from respondents, note that other similar user-input cards like Suggestions also do not show any user data in the agenda overview. Therefore, this card has been updated to only list the poll options that were provided by the admin, so that they can see the card and edit if needed.

### DIFF
--- a/client/lib/features/events/features/event_page/presentation/widgets/survey_answer_tile.dart
+++ b/client/lib/features/events/features/event_page/presentation/widgets/survey_answer_tile.dart
@@ -16,65 +16,80 @@ class SurveyAnswerTile extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     const kIndicatorHeight = 5.0;
+    final showOptionData =
+        answeredParticipants != null && totalParticipants != null;
 
     return Column(
       crossAxisAlignment: CrossAxisAlignment.start,
       mainAxisSize: MainAxisSize.min,
       children: [
         SizedBox(height: 10),
-        Text(
-          answer,
-          style: context.theme.textTheme.bodyMedium!
-              .copyWith(color: context.theme.colorScheme.onSurfaceVariant),
-        ),
         Row(
           children: [
-            Expanded(
-              child: LayoutBuilder(
-                builder: (context, constraints) {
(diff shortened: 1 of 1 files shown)


epenn (approved):
* Created an event with a poll and observed the new card with the new styling hides the previous 0/0.
* And definitely appreciate the new assertion on the constructor.
* LGTM!

PR #313, opened 2026-02-25
fix(chat): prevent display-name flicker and admin-badge duplication on hover

### LINKED ISSUE #248: [Bug] Meeting chat temporarily hides user name, duplicates admin label in previously sent message

1. Cursor moved over chat panel
2. User name disappeared and admin label duplicated

https://github.com/user-attachments/assets/8afec293-d257-4308-bc90-461c4e75fc25

### Expected behavior

Username should always be visible, no duplication of admin tag

### Actual behavior

eeting chat temporarily hides user name, duplicates admin label in previously sent message

### Additional details

_No response_

Firefox 144.0

### Operating system

macOS 15.7.1

novavale:
## Problem

Fixes #248

When hovering over a chat message, two visual bugs appeared:
1. **Display name disappeared** — temporarily replaced by `...`
2. **Admin/mod badge duplicated** momentarily in the header row

## Root Cause

`MessageDisplayState` stored hover state as a `bool` field and called `setState()` on enter/exit. This triggered a full rebuild of the widget, including `UserInfoBuilder`. During that rebuild `UserInfoBuilder` briefly re-entered its loading state, making `snapshot.data == null` and causing the display name to fall back to `'...'`. The same rebuild caused `Wrap` to re-lay out its children, producing a momentary duplicate paint of the admin badge.

## Fix

Replace the `bool _isHovered` field with a `ValueNotifier<bool>` and wrap only the background `Container` in a `ValueListenableBuilder`. The message content `Row` is passed as the static `child` argument of `ValueListenableBuilder` — Flutter reuses it unchanged when the notifier fires, so `UserInfoBuilder` is never re-triggered by a hover event.

```dart
// Before — setState triggers full rebuild
onEnter: (hover) => setState(() => _isHovered = true),

// After — only Container color re-evaluates
final _isHovered = ValueNotifier<bool>(false);
onEnter: (_) => _isHovered.value = true,
```

## Testing

- Hovered over chat messages with admin/mod senders — name stays stable, no badge duplication
- Verified hover highlight still appears/disappears correctly
- Verified delete button still renders for permitted users

### DIFF
--- a/client/lib/features/chat/presentation/widgets/message_display.dart
+++ b/client/lib/features/chat/presentation/widgets/message_display.dart
@@ -34,7 +34,17 @@ class MessageDisplay extends StatefulWidget {
 }
 
 class MessageDisplayState extends State<MessageDisplay> {
-  bool _isHovered = false;
+  // Use ValueNotifier so hover-state changes only rebuild the Container's
+  // color decoration, not the entire widget tree (which would cause
+  // UserInfoBuilder to reload and briefly show '...' for the display name
+  // and cause the admin badge to flicker/duplicate in the Wrap layout).
+  final _isHovered = ValueNotifier<bool>(false);
+
+  @override
+  void dispose() {
+    _isHovered.dispose();
+    super.dispose();
+  }
 
   bool get _canDelete =>
       EventPermissionsProvider.watch(context)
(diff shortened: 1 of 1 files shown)


mikewillems (approved):
Great job, thanks for the PR!

PR #318, opened 2026-03-03
store and calculate both server and client side breakout timestamps in UTC

karenunify:
## What is in this PR?
This PR fixes the "Breakout room matching starting in XX:XX" message and prevents the countdown from resetting to an hour once the breakoutRoomRemainingTime countdown reaches 00:00, as it was doing previously.

## Changes in the codebase
This converts the storing of hostless meeting breakout logic timestamps on both server and client side to UTC. Now the calculation of breakoutRoomRemainingTime and breakoutRoomScheduledTime countdown logic and display avoids timezone mismatch and erroneous display of remaining time until participants are sent into breakout rooms.

* 

## Testing this PR
To test this PR, I created 4 hostless meetings:

1. with no intro or buffer 
2. a buffer but no intro
3. an intro but no buffer
4. both intro and buffer 

In all 4 meeting variations, once the breakoutRoomStatus is in the pending state and the breakoutRoomRemainingTime reaches 0, the breakoutRoomRemainingTime showCountdown should start at 30 seconds consistently, and once it reaches 00:00 should show the "Generating breakout room assignments" breakoutsMessage correctly.


karenunify:
@mikewillems done!

Issue #9, opened 2025-01-10
Breakout Room Recordings


saihla:
A/V recording and easy export thereof must be available for breakout rooms.

## Requirements

1. Recording on/off in event settings for moderator + facilitator + admin
2. Participant notification about recording enabled (like in hosted).
3. Works for all event types (hosted, hostless, livestream)
4. Permissions Gated: Ideally, facilitators and moderators within a community will be able to export recordings for events that they created or attended. Admins should be able to export recordings for all events within a community. These permissions should align with which roles can export other event data.

## Proposed Plan

### Non-technical Summary

Basic Steps:

1. data model for tx and rx
2. app logic for tx and rx
3. tx conversion
4. UI for both
5. connecting to Agora = rx and basic tx done
6. basic captions UI = basic captions done
7. possible real-time sensemaking plugin API and/or (I think really important) consent engine for recording

---

### Phase 1: RecordingSession data model

#### Pitch

Every subsequent phase needs a shared, typed representation of a recording session and its transcript segments. This phase creates that foundation - strictly additive, no functional changes anywhere.

#### Implementation Approach

Scope: `data_models` package only.

1. Add a `RecordingSession` Freezed class in `data_models/lib/recording/recording_session.dart` with the fields described above (`status` and `roomType` as enums, `artifactPaths` as `Map<String, String>`).
2. Add a `TranscriptSegment` Freezed class in the same file.
3. Add constants for the collection name (`recording-sessions`), status and roomType enum values, and `artifactPaths` keys (`hls_playlist`, `complete_mp4`, `transcript_json`).

---

### Phase 2: Explicit recording lifecycle state (start, stop, session persistence)

#### Pitch

Right now recordings start but are never explicitly stopped, the Agora `resourceId` and `sid` are discarded immediately, and there's no record of what was recorded or where it lives. We gain reliable start/stop control, full recoverability, and a persistent session document that everything downstream (video download, transcripts, admin UI) hangs off of. This is the beefiest phase.

#### Implementation Approach

Scope: `agora_api.dart`, `live_meeting_utils.dart`, event-ended flow. Requires Phase 1.

1. Update `recordRoom()` to accept a pre-generated `sessionId`. Write the session document with status `starting` before calling Agora; update with `agoraResourceId` after `acquire`, then `agoraSid` + `gcsPrefix` + status `recording` after `start`. On any failure set status `failed` and write `errorMessage`. ADT preferred for status (we're already on Freezed).
[...]

PR #319, opened 2026-03-04
Fix broken download for long hosted-meeting recordings

### LINKED ISSUE #139: Incorrect "matching starting in..." message

1. Join a hostless event
2. Wait for breakout rooms to begin

Image

### Expected behavior

"Breakout room matching starting in xx:xx" counts down to 0 and then you enter a breakout room.

### Actual behavior

Timer counts down to 0, then starts counting down from 59:59 for a few seconds before starting breakouts.

### Additional details

Breakout rooms do start normally - it doesn't actually set the clock for an hour. But it is a bit concerning as a participant to have it briefly look like it's gonna take an hour.

I tested on Brave as an admin & Chrome as a participant

### Operating system

macOS 14.4.1

mikewillems:
## What is in this PR?

Downloading recordings for long events either spun indefinitely or returned a user-facing error. 

There were two potential causes, this PR fixes both. 

Cause 1: The function was trying to do download by streaming synchronously (bless its heart). 
Cause 2: Agora splits video downloads into a max size of the lesser of 2GB or 2 hours. This could leave a synchronous function hanging until the other download chunk is requested.

More technically:

`downloadRecording` streamed the entire ZIP file through the Cloud Function HTTP response, and the Flutter client read the full response body into memory before creating a blob URL. For long recordings the function hit its timeout mid-stream, the connection drops, and `response.bodyBytes` is null. Normally one would make the function async at that point, but instead the timeout was overridden with 5 minutes instead of 1 minute, which aside from being the wrong approach, still isn't long enough for long recordings, especially depending on network connection. I also don't know what kind of egress bandwidth we have on GCS.

The CORS error visible in devtools was just the signal that we didn't have a file to download.

## Changes in the codebase

Replace ZIP streaming with signed GCS URLs.

The function now lists all MP4 files under `{event.id}/` in GCS, generates a short-lived signed URL for each, and returns them as JSON. File bytes never pass through the function - the whole call returns near-instantly regardless of recording length. The Flutter client renders each URL as a clickable link; the browser handles the download when the user taps it.

### Commit 1 (caea101fb): switching download scheme to signed direct-from-GCS URL

`firebase/functions/js/download-recordings.js`

- Removed ZIP streaming via `archiver` and the 300-second timeout override.
- List MP4 files under `{event.id}/`, generate signed GCS URLs, return as `{ recordings: [{ name, url }] }`.
- Errors now return structured JSON instead of throwing `HttpsError` inside an `onRequest` handler (which does not propagate correctly outside `onCall`).

### Commit 2 (bb807125): hygiene

`firebase/functions/package.json`

Removed unused `archiver` dependency
 
### Commit 3 (6785311c): client download initiation

`client/lib/features/admin/presentation/views/events_tab.dart`

- Parse JSON response and open each signed URL via `html.window.open()`.
- Removed blob creation, anchor click, and URL revocation.
- Added `dart:convert` import.

### Commit 4 (e7a1fe91): hardening the download endpoint

`firebase/functions/js/download-recordings.js`

- Shortened signed URL TTL from 24h to 15 minutes. 24h is longer than necessary and widens the window for a leaked URL to be abused; 15 minutes is long enough for a user to click the link but short enough to meaningfully limit exposure.
- Added `eventPathRegex` validation before passing `eventPath` to `firestore.doc()`. Without it any caller with a valid auth token could supply an arbitrary Firestore path and read documents outside the events collection. The regex enforces the expected shape: `community/{id}/templates/{id}/events/{id}` (derived from the `FirestoreDatabase.templateReference` path builder and `FirestoreEventService.eventsCollection`).
- Sort `mp4Files` by name before generating URLs so recordings are returned in a deterministic order regardless of GCS list ordering.

### Commit 5 (e7df3f37): UI visual consistency, use tapped links instead of popups in case blocked

`client/lib/features/admin/presentation/views/events_tab.dart`

- Replaced `html.window.open()` loop (called after an `async` gap, so popup blockers fire) with a stateful approach: the button press stores the returned URLs in `_recordingLinks` (a `Map<String, List<Map<String, String>>>` keyed by event ID), and `_buildRecordingSection` renders numbered `GestureDetector` links below the button once populated. Links persist on screen until the next navigation, and any subsequent button press refreshes them.
- Added guards around `jsonDecode` (try/catch) and the `recordings` field (`is! List` check) to surface a clean user-facing error instead of a runtime cast exception if the server returns malformed JSON or an unexpected schema.
- Swapped bare `Text` for `HeightConstrainedText` on the rendered links to match every other text element in the table, and removed the hardcoded `fontSize: 12` (nothing in `AppTextStyle` uses 12; the smallest defined size is `bodySmall` at 14). Both changes bring the link style in line with the date link in the same row and other admin-view link patterns (`Colors.blue` + `TextDecoration.underline`).

None.

### For existing recordings (this should test it completely)
1. Go to any long recording over ~1h and attempt to download it.
2. Verify that the app doesn't freeze and that a new tab opens with the proper download.
3. Do the same for a recording over ~2h (maybe 3?) and verify that you get two downloads.

1. Start a hosted meeting with recording, keep it short.
2. Download recording - verify it still works for short recordings, but opens in a new tab.
4. Start a meeting with recording, let it run for 1h.
5. End the meeting and download the recording. Verify that the app continues running and that a new tab opens.
6. Start a meeting with recording, let it run for at least 3h. Verify that the app continues running, and TWO tabs open with two downloads (at this point Agora should split the according into two files per its [docs](https://docs.agora.io/en/cloud-recording/develop/manage-files)).

## Additional information

Note: This fixes downloads for the existing main-room hosted recording flow only. Per-session and breakout room download support is tracked in phases 4 and 5 of [Breakout Room Recordings #9](https://github.com/berkmancenter/frankly/issues/9).

## Status Update as of 19 Mar

Deviated for a while to other issues and merged a bunch of other fixes, so could be a little tough to follow the commit trail. Here's the current state and a guide to the commits:

### Recording-Specific Commits (oldest to newest)

| Commit     | Summary                                                                      | Files                    |
| ---------- | ---------------------------------------------------------------------------- | ------------------------ |
| `caea101f` | Switch from streaming zip to returning signed GCS URLs                       | `download-recordings.js` |
| `bb807125` | Remove unused `archiver` import from `package.json`                          | `package.json`           |
| `6785311c` | Client: pass auth token in header, open signed URLs in new tab               | `events_tab.dart`        |
| `e7a1fe91` | Harden endpoint: shorter TTL (15 min), path validation regex, sorted results | `download-recordings.js` |
| `e7df3f37` | UI: replace pop-up windows with tap-to-open links, guard JSON parsing        | `events_tab.dart`        |
| `31be640b` | Add debug info to 404 error responses                                        | `download-recordings.js` |

#### Backend (`download-recordings.js`)

- **Signed URLs instead of zip streaming**: The original function bundled recordings into a zip archive via `archiver` and streamed it through the HTTP response. This was replaced with generating short-lived (15-min) signed GCS URLs and returning them as JSON.
- **Proper HTTP error responses**: Changed from throwing `HttpsError` (callable-style errors) to returning proper REST responses (`res.status(4xx).json({...})`).
- **Path validation**: Added a regex check (`eventPathRegex`) to validate `eventPath` format before using it in a Firestore query.
- **Sorted results**: MP4 files are now sorted by name before returning.
- **Debug info**: 404 responses include debug context (`eventPath`, `eventId`, `bucket`).
- **Removed `archiver`** dependency from `package.json`.

#### Client (`events_tab.dart`)

- **Auth token in header**: Switched from passing auth in the body to `Authorization: Bearer $idToken` header.
- **Signed URL display**: After clicking Download, the returned URLs are stored in `_recordingLinks` (keyed by event ID) and rendered as clickable "Recording 1", "Recording 2", etc. links below the Download button.
- **GestureDetector instead of window.open**: Uses `GestureDetector.onTap` to open URLs
  to avoid pop-up blockers (the old code called `html.window.open` in a loop). 
- **Guarded JSON parsing**: `jsonDecode` is wrapped in a try/catch, and `rawList` is type-checked before use.

#### 1. Download button errors when recordings aren't ready

The Download button calls the `downloadRecording` function immediately. If the recording hasn't finished processing yet (no MP4 files in the bucket), the function returns a 404 with `"No recordings found"`, which surfaces as an error alert to the user. There is no handling for the "not ready yet" case - no polling, no graceful message, and no disabled state.

#### 2. Links are not shown until user clicks Download

The intended behavior (per user recollection) was to show the download links section proactively, with links disabled/greyed until the recording is available. Currently, the links only appear *after* a successful Download button click. There is no automatic polling or readiness check on page load.

#### 3. No tooltip/alt text for "still processing" state

Add tooltip text explaining that recordings are still being prepared. Not yet implemented.

#### 4. Download button is always enabled

The button doesn't check whether recordings exist before allowing a click. It could be disabled or hidden when recordings aren't expected to be ready, or it could show a different state.

#### 5. Debug code still present

Commit `31be640b` adds `debug` fields to 404 error responses. This is useful for development but may want to be gated or removed before merging to staging.

#### 6. Error message is generic

The client throws `Exception('Failed to get recording URLs')` for any non-200 response, discarding the server's specific error message. (This was partially addressed in the `mw/dev/fix-local-recordings-setup` branch with a debug-gated improvement.)

### Non-Recording Changes Also on This Branch

The branch also includes unrelated commits merged from staging:

- `477b3644` - Revert deletion of `.enterMeeting()` call in `_joinEvent()`
- `f3732fdc` - Regenerated `package-lock.json`
- Various i18n, discussion thread, chat, and community changes from upstream merges

These are already in staging and shouldn't conflict.

### DIFF
--- a/client/lib/features/admin/presentation/views/events_tab.dart
+++ b/client/lib/features/admin/presentation/views/events_tab.dart
@@ -30,6 +30,9 @@ class _EventsTabState extends State<EventsTab> {
 
   var _numToShow = 10;
 
+  /// Signed GCS URLs returned for each event, keyed by event ID.
+  final Map<String, List<Map<String, String>>> _recordingLinks = {};
+
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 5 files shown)


katherineqian (commented):
Going to just submit the comments I have at this stage - I completed a review of the code and it looks good to me. Still need to re-test the behavior and the UI; please re-request when ready!

mikewillems:
Closed because superceded by hosted recordings fix "v2" PR.

PR #18, opened 2025-01-18
Add live meeting tests

jencompgeek:
## What is in this PR?
Adds tests for the remaining live meeting functions

## Changes in the codebase
New tests, also slight improvement to test data cleanup to delete some orphaned collectionGroups

## Changes outside the codebase
None

## Testing this PR
Execute tests

## Additional information
I considered extracting creation of a test event to the util class, but not sure it's worth the ROI since pretty much all of the function tests have already been written and will be rewritten in the near future.

PR #14, opened 2025-01-15
Update Flutter installation section of README for clarity + specify Flutter version.

katherineqian:
## What is in this PR?

To avoid unintended errors in development, we should specify that Flutter 3.22.2 is officially supported in the README. I added a note about this and along the way, fixed the Markdown formatting + cleaned up the Flutter installation section so it should be significantly more readable than before.

## Changes in the codebase

* README updates.

## Changes outside the codebase

* N/A

## Testing this PR

 N/A

## Additional information

N/A

Issue #13, opened 2025-01-11
Convert functions to TypeScript


saihla:
Convert functions to TypeScript & remove dart-to-js build step

# Acceptance Criteria

- [ ]  Functions are in Typescript
- [ ]  Dart-to-js build step removed

Issue #12, opened 2025-01-11
Internationalization


saihla:
Enable easy translation of all text components to support multiple languages and broaden global accessibility.

- [x] adopt Flutter globalization framework
- [x] replace english strings with variables
- [x] provide minimal Spanish translation (other languages later)
- [ ] Verify Spanish translations are accurate

PR #337, opened 2026-03-12
Fix emulator issues and browser bugs in local dev, simplify launch

mikewillems:
# PR: Fix and simplify local dev

## Summary

Before this PR, running the emulators locally required manually running `dart run build_runner build --output=build` and then `firebase emulators:start` with the right flags, from the right directory, remembering to pass `--project dev`. The emulator startup script (`emulators.sh`) was two lines: a build command and a hardcoded `firebase emulators:start` with `--project dev` and `--inspect-functions` always on. There was no way to start everything from one command, no port conflict detection, no graceful shutdown, and no incremental builds. The Firestore emulator port was misconfigured (8081 instead of 8080). If the emulators crashed or got stuck, the only option was to hunt for orphan processes manually.

On the client side, the emulator was essentially unusable. The RTDB emulator connection was wired up too late (inside `UserDataService.initialize()`, after services had already touched the RTDB instance), producing a fatal "already initialized" crash on every startup. Email sign-in fired `_handleUserSignedIn` twice (once explicitly, once via the auth state stream), causing duplicate Firestore writes. Emulator users (who never have `displayName` on their auth record) all appeared as "User-ab12" because the display name fallback skipped the email entirely. The `authMessageOnError` helper crashed on null `callback` and silently swallowed unknown exception types. Cloud functions crashed on any model with `required String id` because Firestore doesn't store document IDs as fields, and `getFirestoreObject` never injected them. The `joinEvent` function crashed with a null deref when the participant doc didn't exist yet. And tapping "Enter Event" joined the user but never transitioned to the meeting UI because `_joinEvent` was missing the `enterMeeting()` call.

This PR adds a one-command `npm run dev` workflow with incremental builds and port management, fixes all of the above client and function bugs, and rewrites `docs/pages/dev.md` to document the new setup.

---

### New: one-command dev startup

- `run-dev.sh` orchestrates the full local startup: pub get, npm install, data_models build, functions build, emulator start, Flutter client launch
- `npm run dev` wired in both root and `firebase/functions` `package.json`
- Stamp files in `.local/dev-stamps/` enable incremental builds (skip unchanged steps)

### New: emulator lifecycle scripts

- `emulators-start.sh` checks for port conflicts before launching, pointing to `emulators-stop.sh` if ports are in use
- `emulators-stop.sh` does a graceful shutdown via the emulator hub API, with escalating fallbacks: SIGTERM on listener PIDs, then SIGKILL, with clear diagnostics if something is stuck
- `npm run emulators:stop` and `npm run emulators:reset` convenience scripts

### Emulator configuration fixes

- `emulators.sh` reads `FIREBASE_PROJECT_ID` from `client/.env` instead of hardcoding `--project dev` (mismatch causes CORS errors with no helpful error message)
- `--inspect-functions` is now opt-in via `FRANKLY_DEBUG_FUNCTIONS=1` (was always on, forcing sequential/slow execution)
- Fail-fast on missing `.env` or unset project ID
- `SKIP_DART_BUILD=1` prevents double-building when called from `run-dev.sh`
- Firestore emulator port corrected from 8081 to 8080 in `firebase.json`

### Client crash fixes

- `app.dart`: moved `useDatabaseEmulator` to immediately after `Firebase.initializeApp` (was in `UserDataService.initialize()`, too late)
- `user_service.dart`: removed duplicate `_handleUserSignedIn` call in `signInWithEmail`
- `user_service.dart`: fixed display name fallback chain to include email-derived name
- `user_service.dart`: replaced `currentUserId!` with captured IDs after `await` points
- `error_utils.dart`: `callback!()` -> `callback?.call()` (was crashing on null optional)
- `error_utils.dart`: added `else` branch for non-Firebase exceptions (were silently swallowed)

### Cloud function fixes

- `firestore_utils.dart`: `getFirestoreObject` now injects `map['id']` from the document path
- `join_event.dart`: explicit `exists` check before accessing participant doc data

### Event page fixes

- `event_page.dart`: `_joinEvent` now calls `enterMeeting()` after a successful join, completing the transition to the meeting UI
- `event_page.dart`: simplified `_joinEvent` return type and removed dead conditional toast
- `event_info.dart`: removed unreachable "past concluded event" branch and redundant "not entered" toast

### Documentation

- `docs/pages/dev.md`: reformatted from tabbed layout to flat markdown, added "Running Everything with One Command" section, env var table, "Running Emulators Only" section, emulator recovery docs, dynamic project ID instructions
- `client/.env.local.example`: added `FIREBASE_PROJECT_ID` placeholder
- `client/.gitignore`: added FlutterFire CLI sidecar `firebase.json`
- Fixed `.env.example.local` -> `.env.local.example` filename in docs

---

### #1 `9cdd856` -- dev.md formatting

Reformats `docs/pages/dev.md` from the mkdocs tabbed layout (indented content under `=== "macOS"` / `=== "Linux"` tabs) to flat markdown. No content changes.

### #2 `cc4cc23` -- emulator start/stop scripts

Adds `emulators-start.sh` (port conflict check before launch) and `emulators-stop.sh` (graceful shutdown with escalating fallbacks). Also adds a shebang and `set -euo pipefail` to `emulators.sh`, which previously had neither.

### #3 `8775715` -- dev.md emulator section rewrite

Expands the emulator docs in `dev.md` with port listing, recovery instructions for stuck emulators, and references to the new start/stop scripts.

### #4 `6d1443c` -- firestore port fix

The Firestore emulator was configured on port 8081 in `firebase.json` but the client expected 8080. Corrected to 8080. The rest of the diff is whitespace formatting.

### #5 `9db3ce3` -- add missing port to start check

Adds a port that was missing from `emulators-start.sh`'s conflict check and fixes a minor quoting issue in `emulators-stop.sh`.

### #6 `1811522` -- dev.md port description fix

One-line fix to a port description in the docs.

### #7 `bac439e` -- package.json script update and dependency cleanup

Points `npm run emulators` at `emulators-start.sh` instead of `emulators.sh` directly (so the port check runs first). Removes `firebase-tools` from devDependencies (it's installed globally). The `package-lock.json` diff is large but it's just the lockfile regenerating after the devDependency removal.

### #8 `ee3cd58` -- `npm run dev` startup script

Adds `run-dev.sh` and wires `npm run dev` in both `package.json` files. The script orchestrates the full local startup with incremental builds using stamp files. Also adds `emulators:stop` and `emulators:reset` convenience scripts.

### #9 `f48785f` -- stamp files for incremental builds

Adds `.local/dev-stamps/` with empty stamp files that `run-dev.sh` touches after each successful build step. The `need_data_models_build` and `need_functions_build` functions compare source file timestamps against the stamp.

### #10 `24192f0` -- gitignore FlutterFire CLI sidecar

The FlutterFire CLI generates `client/firebase.json` with a developer-specific project ID. It was never committed (the file itself is not in the repo), but it was perpetually showing as an untracked file in `git status`. Adding it to `.gitignore` silences that noise.

### #11 `bcdb388` -- dynamic project ID and emulator flags

The emulator script used `--project dev`, but the client reads the actual project ID from `client/.env`. If these don't match, every Firebase call fails with CORS errors -- and the error says nothing about project ID, making it very hard to diagnose. This commit reads `FIREBASE_PROJECT_ID` from `client/.env`, fails fast if it's missing, and makes `--inspect-functions` opt-in. The bash 3.2 array expansion pattern (`${inspect_flag:+"${inspect_flag[@]}"}`) is a macOS-specific workaround: Apple ships bash 3.2 as `/bin/bash` (frozen at GPLv2). In bash 3.2 with `set -u`, referencing an empty array as `"${arr[@]}"` raises "unbound variable". Bash 4.x/5.x (standard on Linux) handles empty arrays cleanly, so the pattern is only needed for macOS compatibility. It's harmless on bash 4+.

### #12 `9327abe` -- fix build detection and port wait in run-dev.sh

Two bugs in the new `run-dev.sh` from #8. The `find` commands referenced `bin/` and `test/` directories that don't exist in `data_models/`, causing them to exit non-zero. Under `set -euo pipefail` this made the build-needed check silently return false, so incremental builds never triggered. Also, `wait_for_port` used `$(date +%s)` in arithmetic, which bash 3.2 can't handle. Replaced with the `$SECONDS` builtin.

### #13 `d2bf7cb` -- inject document ID in getFirestoreObject

**NOTE: RELOCATED TO SEPARATE PR** (`mw/fix/firestore-jsonnull-fix`) - This commit addresses JSNull deserialization issues unrelated to local DX improvements.

Firestore doesn't store the document ID as a field. Models with `required String id` (Template, Community, etc.) crash with `JSNull is not a subtype of String` on deserialization. The fix sets `map['id'] ??= path.split('/').last` before calling the constructor. Uses `??=` rather than `putIfAbsent` because the serializer writes `id: null` -- the key exists but is null, so `putIfAbsent` would skip it.

### #14 `44d9892` -- auth and user service fixes

Four related fixes:

1. **RTDB init order**: `useDatabaseEmulator` was called inside `UserDataService.initialize()`, which runs after services already have RTDB references. Firebase throws a fatal error at that point. Moved to `app.dart` immediately after `Firebase.initializeApp`.

2. **Double sign-in handler**: `signInWithEmail` explicitly called `_handleUserSignedIn` after `signInWithEmailAndPassword`, but `authStateChanges` fires it automatically. This caused duplicate `createCurrentUserInfoIfNotExists` writes.

3. **Display name fallback** (developer experience): Was `auth.displayName ?? 'User-${uid.substring(0, 4)}'`, entirely skipping email. Emulator users never have `displayName` (auth emulator doesn't set it), so all emulator test accounts appeared as "User-ab12" instead of identifiable names. Not a prod bug. New chain: explicit arg -> `auth.displayName` -> email-derived initials -> uid prefix. Makes local testing more readable when you have multiple test accounts.

4. **Post-await null safety**: `currentUserId!` after `await` points throws if the user signs out during the await. Changed to capture `userInfo.id` before the await.

### #15 `d913438` -- error_utils crash fixes

`authMessageOnError` had `callback!()` on an optional parameter -- crashes when null. Changed to `callback?.call()`. Also added an `else` branch for non-Firebase/non-Visible exceptions, which were being caught and silently dropped.

### #16 `72f24ff` -- wire enterMeeting after join

The `enterMeeting()` call was accidentally dropped from `_joinEvent` in `6d2119b6` ("Show error toast when users fail to join") and restored via PR #322 (`kq/fix/fix-join-event-button`), which is the PR base for this branch. This commit refactors the same method to a cleaner shape: removes the nested `alertOnError` wrapping (the outer wrapper was returning `false` on success due to how Dart's null coalescing interacts with `alertOnError`'s return type), removes the dead conditional toast in `event_info.dart`, and removes an unreachable code path for past concluded events.

### #17 `5c15ebc` -- env example and project ID docs

Adds `FIREBASE_PROJECT_ID=<your-firebase-project-id>` to `client/.env.local.example`. Updates `docs/pages/dev.md` to reference the dynamic project ID instead of hardcoded `dev`.

### #18 `6d7ae91` -- document the new dev workflow

Adds "Running Everything with One Command" to `docs/pages/dev.md`: what `npm run dev` does, the stamp-file mechanism, and environment variable table for `SKIP_DART_BUILD` and `FRANKLY_DEBUG_FUNCTIONS`. Separates `npm run emulators` into its own "Running Emulators Only" section. Fixes `.env.example.local` -> `.env.local.example`.

### #19 `f4b81b2` -- revert functions file watcher ignore

Reverts the `functions[0].ignore` in `firebase.json` back to the default list (`node_modules`, `.git`, debug logs, `*.local`). The previous value (`["**"]`) disabled the emulator's chokidar file watcher but also broke `firebase deploy --only functions` -- the deploy packaging code (`prepareFunctionsUpload.js`) uses the same ignore list to build the upload zip, so `**` would produce an empty deployment. The original symptom (functions reloading on `git checkout`) was caused by orphaned emulator processes from previous `run-dev.sh` invocations, not by the watcher itself.

---

## Production impact assessment

**UPDATE: Firestore and join_event fixes below are now in a separate PR** (`mw/fix/firestore-jsonnull-fix`). Only sections #3-9 below apply to this PR.

### 1. `firestore_utils.dart` -- document ID injection (commit #13)

**Change:** `map['id'] ??= path.split('/').last` added to `getFirestoreObject`, which is called ~40 times across all cloud functions.

**Risk: Verified Low** -- broadest potential impact of commits in this PR (40 call sites), but confirmed safe after reviewing all model definitions and generated `fromJson` code.

**Models with `required String id`** (the fix is necessary for these -- they crash without it): `Event`, `Community`, `Participant`, `Template`, `MeetingGuide`/`MeetingUserSuggestion`. In every case, `id` means "Firestore document ID" and nothing else. The serializer (`_$$_XToJson`) writes `id` to the map, but Firestore doesn't persist it, so `fromJson` receives `id: null`. The `??=` correctly fills it from the document path.

**Models without an `id` field** (the fix is harmless for these -- extra key is ignored): `LiveMeeting` (uses `meetingId`), `BreakoutRoom` (uses `roomId`), `Membership` (uses `userId` + `communityId`). Their generated `_$$_XFromJson` functions only read their own named keys; the `map['id']` entry set by `??=` is silently discarded by json_serializable.

No model uses `id` for anything other than the Firestore document ID.

**Benefit:** Without this fix, _every_ function that reads a model with `required String id` crashes with `JSNull is not a subtype of String`. This is a blocking bug for emulators and would also crash in prod if any function path hits a doc that doesn't have `id` written explicitly.

**How to verify:** Deploy to a staging project. Call `joinEvent`, `createEvent`, `getMeetingJoinInfo`, and `resolveJoinRequest`. Confirm each returns successfully and the response includes the correct `id`. Check function logs for any unexpected `id` values.

### 2. `join_event.dart` -- participant exists guard (commit #13)

**Change:** Replaced `getFirestoreObject` (which unconditionally calls `.data.toMap()`) with `get()` + `exists` check.

**Risk: Low** The only behavioral change is that a missing participant doc now returns cleanly instead of crashing. No existing successful path is altered.

**Benefit:** Prevents 500 errors when a user calls `joinEvent` before the participant doc exists (e.g., fresh emulator, or a race condition in prod).

**How to verify:** Call `joinEvent` for a user who hasn't joined yet. Confirm no 500 error.

### 3. `user_service.dart` -- display name fallback chain (commit #14)

**Change:** `firstAndLastInitial(displayName ?? auth.displayName) ?? 'User-${uid}'` replaced with `firstAndLastInitial(displayName) ?? firstAndLastInitial(auth.displayName) ?? firstAndLastInitial(auth.email) ?? 'User-${uid}'`.

**Risk: Very Low.** This is a developer experience fix, not a prod bug fix. In prod, sign-in via Google or registration via email both write `displayName`, so the new fallback chain only applies to emulator test accounts (which never have `displayName` set). For any prod user with `displayName` set, the behavior is identical to before.

**Benefit:** Emulator test accounts show email-derived initials instead of all being "User-ab12", making local testing readable when you have multiple accounts.

**How to verify:** Sign in on staging. Confirm display name is unchanged for existing accounts.

### 4. `user_service.dart` -- removed explicit `_handleUserSignedIn` in `signInWithEmail` (commit #14)

**Change:** Removed `await _handleUserSignedIn(user)` after `signInWithEmailAndPassword`. The `authStateChanges` listener fires `_handleUserSignedIn` automatically.

**Risk: Low** `authStateChanges` is guaranteed by Firebase Auth to fire after every sign-in on all platforms. Verified that every other sign-in method in `user_service.dart` already relies on the same pattern:

- `signInAnonymously()` (line 246) does not call `_handleUserSignedIn` - the listener does
- `registerWithEmail()` (line 281) stores `_emailRegistrationDisplayName` specifically _for the listener to use_ (comment on line 284 says so explicitly)
- `signInWithGoogle()` (line 272) uses redirect - the listener handles it on return

The listener is set up in `initialize()` (line 129), which runs before any sign-in is possible. The old explicit `_handleUserSignedIn` call was the only sign-in method that did this - it was the inconsistency, not the norm.

**Benefit:** Prevents double execution of `createCurrentUserInfoIfNotExists`, which writes duplicate Firestore documents.

**How to verify:** Sign in with email on staging. Confirm the user document is created once (check Firestore writes in logs or the Firestore console for duplicates).

### 5. `user_service.dart` -- post-await ID safety (commit #14)

**Change:** `currentUserId!` after `await` replaced with `userInfo.id` / `newUserInfo.id`.

**Risk: Very Low** The old code would crash if the user signed out during an ongoing `await`. The new code uses a local variable captured before the `await`, which is always safe.

**Benefit:** Eliminates a potential null-deref crash.

### 6. `error_utils.dart` -- `callback?.call()` (commit #15)

**Change:** `callback!()` changed to `callback?.call()`.

**Risk: Very Low** 2 of 3 call sites don't pass `callback`, so they were crashing on success. This just stops the crash.

**Benefit:** `authMessageOnError` no longer throws on happy-path when `callback` is omitted.

### 7. `error_utils.dart` -- `else` branch for unknown exceptions (commit #15)

**Change:** Non-`FirebaseAuthException`, non-`VisibleException` errors now call `errorCallback(sanitizedError, 'unknown')` instead of being silently caught and dropped.

**Risk: Low** `_accountErrorMessageBuilder` has a `default:` branch for all unrecognized codes, which shows `context.l10n.somethingWentWrongTryAgain` - a localized friendly message. Even if the error is still `'unknown'`, the raw string is never rendered to the user.

The only actions wrapped in `authMessageOnError` are `resetPassword` and `_onSubmit` (sign-in/register). Firebase wraps all network and auth errors in `FirebaseAuthException`, so the `else` branch would only fire for something outside Firebase's own error handling - e.g., if `widget.onComplete?.call()` in `_onSubmit` throws after a successful sign-in. In that edge case the user would see "Something went wrong, try again" even though they actually signed in. But this is an improvement over the previous behavior: before, the exception was silently swallowed and the form just sat there with no feedback at all, appearing broken.

**Potential issue:** None beyond the edge case above, which is extremely unlikely in practice.

**Benefit:** Errors are no longer silently swallowed. We can see that something failed.

**How to verify:** Trigger a non-Firebase auth error on staging (e.g., network timeout). Confirm the error message is reasonable, not raw "unknown".

### 8. `event_page.dart` -- enterMeeting after join (commit #16)

**Change:** `_joinEvent` now calls `eventPageProvider.enterMeeting()` after a successful join when `enterMeeting: true`. Also removed the outer `alertOnError` wrapper (replaced with a single layer), removed dead code in `event_info.dart`.

**Risk: Low** This fixes a clearly broken flow -- without `enterMeeting()`, the user joins but the UI never transitions. The only concern is whether any code path relied on `_joinEvent` _not_ entering the meeting (e.g., join-only without entering). The `enterMeeting` parameter guards this.

**Benefit:** "Enter Event" button actually works.

**How to verify:** Join and enter an event on staging. Confirm the meeting dialog appears.

### 9. `app.dart` + `user_data_service.dart` -- RTDB emulator init order (commit #14)

**Change:** `useDatabaseEmulator` moved from `UserDataService.initialize()` to `app.dart`, gated on `UserDataService.usingEmulator`.

**Risk: None for prod.** The `usingEmulator` flag is only set in the emulator entry point (`dev_emulators_main.dart`). In production builds, `usingEmulator` is `false` and the new code in `app.dart` is never reached. The removed call in `user_data_service.dart` was also gated on the same flag.

### 10. `firebase/functions/package.json` -- `firebase-tools` removed from devDependencies (commit #7)

**Risk: None** CI installs `firebase-tools` globally (`npm install -g firebase-tools` in `.github/workflows/test_firebase.yaml:83`). The local devDependency was redundant and version-pinned to an old `^9.18.0`.

---

### 1. Full startup from scratch

```
npm run dev
```

Verify:

- Emulators start and print "All emulators ready"
- Flutter client launches and connects to local emulators
- No CORS errors in the browser console
- Emulator UI is accessible at <http://127.0.0.1:4000>

### 2. Port conflict detection

- Start emulators with `npm run emulators`
- In another terminal, run `npm run emulators` again
- Verify: the second invocation exits immediately with a message about ports already in use, pointing to `emulators-stop.sh`

### 3. Emulator stop and reset

- With emulators running, run `npm run emulators:stop` from `firebase/functions`
- Verify: emulators shut down cleanly ("stopped cleanly" or "stopped after targeted kill")
- Run `npm run emulators:reset` -- verify: stops any leftover processes, then starts fresh

### 4. Sign in with email

- Create a user in the Auth emulator UI or use the client's sign-up flow
- Sign in with that email/password
- Verify: no crash, no duplicate Firestore writes in the emulator logs, display name shows email-derived initials (not "User-xxxx")

### 5. Join and enter an event

- Create a community and event (or use seed data)
- Tap "Enter Event" on the event page
- Verify: the meeting dialog appears after joining (not stuck on the event page)
- Also verify: joining an event that hasn't started shows the "not started yet" alert

### 6. Incremental builds

- Run `npm run dev` (full startup)
- Stop with Ctrl+C
- Run `npm run dev` again without changing any source files
- Verify: output shows builds being skipped ("up to date" or similar)
- Touch a file: `touch data_models/lib/something.dart`
- Run `npm run dev` again
- Verify: data_models rebuilds but functions do not

### 7. Debug mode

```
FRANKLY_DEBUG_FUNCTIONS=1 npm run dev
```

Verify: emulator output includes inspector/debugging message. Without the env var, no inspector output.

### 8. Missing .env / project ID

- Rename `client/.env` temporarily (e.g., `mv client/.env client/.env.bak`)
- Run `npm run emulators`
- Verify: clear error message about missing `.env`
- Restore: `mv client/.env.bak client/.env`

### 9. Regression check -- production Firebase

- Test on the staging deployment preview
- Sign in, browse communities, join an event, enter a meeting
- Verify: no regressions from the `user_service.dart`, `error_utils.dart`, `event_page.dart`, or `firestore_utils.dart` changes

---

## Post Review Additions

The PR was split into two after initial review feedback requested separating unrelated concerns:

### 1. Created Separate JSNull Fix PR: `mw/fix/firestore-jsonnull-fix`

**Rationale**: Commits #13 (`d2bf7cbc` - firestore_utils and join_event fixes) and the firestore-related parts of #14 address backend deserialization issues unrelated to local DX improvements. These should be reviewed and merged independently per @kq's correct feedback.

**Commits in separate PR**:

- `d2bf7cbc`: "fix model lookup errors for fields with 'required String id'"
- `f55c0d89`: "extend doc ID injection fix to JoinEvent participant (missed in getFirestoreObject fix)"

This PR (mw/dev/fix-and-simplify-local-dev) now only ships local DX improvements - see updated production impact sections #3-9.

### 2. Simplified `emulators-stop.sh` Architecture

**Commit**: `b2aec65d` ("Extract port list to simplify emulators-stop.sh")

**Changes**:

- Moved emulator ports from hardcoded lists to a top-level `EMULATOR_PORTS` array
- Created `format_ports_for_lsof()` helper to generate lsof port flags
- Extracted repeated kill logic into `kill_listeners()` function with signal parameter

**Benefit**: Eliminates three duplicate full port lists in the script. Future port changes only need updating one place. Same functionality, cleaner code.

### 3. Removed `firebase-tools` from Package.json Files

**Change**: Removed `firebase-tools` from devDependencies in both `firebase/functions/package.json` and `firebase/firestore/package.json`.

We don't use `firebase-tools` anywhere in the codebase - only through CLI commands like `firebase emulators:start`. It's already installed globally in CI (`.github/workflows/test_firebase.yaml:83`).

### DIFF
--- a/docs/pages/dev.md
+++ b/docs/pages/dev.md
@@ -12,21 +12,21 @@ This section covers setting up a new computer for Flutter development.
 ### Part 1: Platform-specific
 
 === "macOS" 1. Download and install Google Chrome [here](https://www.google.com/chrome/) if it’s not already pre-installed. This is used for live debugging on web. 2. Download and install XCode from the Mac App Store. This is used for developing iOS apps and running on macOS as a desktop app. 3. Optional, but recommended: Install Homebrew [here](https://brew.sh/). 4. Xcode should've installed git automatically, but if not for some reason, you can install it via Homebrew:
-` brew install git
-      ` 5. Clone the Frankly repo in a directory where you prefer your projects to live:
-` git clone https://github.com/berkmancenter/frankly && cd frankly
-      ` 6. Follow the instructions [here](https://docs.flutter.dev/get-started/install) to install Flutter on your machine. You can choose iOS as your target platform. - This includes a link to install CocoaPods. However, you may run into issues with installing CocoaPods due to a Ruby version issue (the pre-installed Ruby on MacOS is too old). You can install ruby via Homebrew instead by running `brew install cocoapods`, which should alleviate those errors. - **Recommended**: Install the Flutter SDK in your home folder under a directory called `dev` (or something similar). 7. Install VSCode [here](https://code.visualstudio.com/download). - **Recommended**: Install the [Flutter VSCode extension](https://docs.flutter.dev/get-started/editor#install-the-vs-code-flutter-extension) and use the extension to [install Flutter via VSCode](https://docs.flutter.dev/get-started/install/macos/mobile-ios#use-vs-code-to-install-flutter). 8. Add Flutter to your PATH. For Mac with Zsh (you can also copy this command from [here](https://docs.flutter.dev/get-started/install/macos/mobile-ios?tab=download#add-flutter-to-your-path)), create or open ~/.zshenv and add this line:
-` export PATH=$HOME/dev/flutter/bin:$PATH
-        `
+`brew install git
+     ` 5. Clone the Frankly repo in a directory where you prefer your projects to live:
+`git clone https://github.com/berkmancenter/frankly && cd frankly
+     ` 6. Follow the instructions [here](https://docs.flutter.dev/get-started/install) to install Flutter on your machine. You can choose iOS as your target platform. - This includes a link to install CocoaPods. However, you may run into issues with installing CocoaPods due to a Ruby version issue (the pre-installed Ruby on MacOS is too old). You can install ruby via Homebrew instead by running `brew install cocoapods`, which should alleviate those errors. - **Recommended**: Install the Flutter SDK in your home folder under a directory called `dev` (or something similar). 7. Install VSCode [here](https://code.visualstudio.com/download). - **Recommended**: Install the [Flutter VSCode extension](https://docs.flutter.dev/get-started/editor#install-the-vs-code-flutter-extension) and use the extension to [install Flutter via VSCode](https://docs.flutter.dev/get-started/install/macos/mobile-ios#use-vs-code-to-install-flutter). 8. Add Flutter to your PATH. For Mac with Zsh (you can also copy this command from [here](https://docs.flutter.dev/get-started/install/macos/mobile-ios?tab=download#add-flutter-to-your-path)), create or open ~/.zshenv and add this line:
+`export PATH=$HOME/dev/flutter/bin:$PATH
+       `
 Restart terminal sessions to see the changes.
 === "Linux" 1. Download and install chromium and git if they're not already installed. Chromium is used for live debugging on web.
-` sudo apt-get update && sudo apt-get upgrade && sudo apt-get install -y chromium git
-      `
+`sudo apt-get update && sudo apt-get upgrade && sudo apt-get install -y chromium git
+     `
 a. Set the path to chromium so flutter can find it:
-` export CHROME_EXECUTABLE=$(which chromium)
-      ` 2. Clone the Frankly repo in a directory where you prefer your projects to live:
-` git clone https://github.com/berkmancenter/frankly && cd frankly
-      ` 3. Follow the instructions [here](https://docs.flutter.dev/get-started/install) to install Flutter dependencies on your machine. You can choose web as your target platform. 4. Install VSCode [here](https://code.visualstudio.com/docs/setup/linux).
+`export CHROME_EXECUTABLE=$(which chromium)
+     ` 2. Clone the Frankly repo in a directory where you prefer your projects to live:
+`git clone https://github.com/berkmancenter/frankly && cd frankly
+     ` 3. Follow the instructions [here](https://docs.flutter.dev/get-started/install) to install Flutter dependencies on your machine. You can choose web as your target platform. 4. Install VSCode [here](https://code.visualstudio.com/docs/setup/linux).
 
         !!! info ""
             You may need to download the binary for your specific architecture [here](https://code.visualstudio.com/Download)
@@ -143,34 +143,154 @@ You don't need to run `npm install` again unless you've added new dependencies o
 
 #### Emulators
 
-Firebase has a full suite of emulators called Firebase Local Emulator Suite. You can find the full description of the Firebase Local Emulator Suite and its capabilities [here](https://firebase.google.com/docs/emulator-suite).
+Frankly uses the **Firebase Local Emulator Suite** for local development. The emulator suite provides local versions of Firebase services: Firestore, Auth, Realtime Database, Pub/Sub, and Cloud Functions.
 
-You should emulate services locally for development purposes, and set up the client to use these emulators instead of connecting to a live Firebase project. By default, the emulators will run against the default project "dev," specified in the `.firebaserc` file.
+Documentation for the emulator suite is available here:  
+<https://firebase.google.com/docs/emulator-suite>
 
-**Using config in emulators**
+By default the emulators run against the Firebase project named **`dev`**, which is specified in `.firebaserc`.
 
-You do not need to run `functions:config:set.` as the emulators are configured by a file.
+---
 
-- To configure the emulators, create the file `firebase/functions/.runtimeconfig.json`.
-  - A sample file containing the config properties described above can be found in `firebase/functions/.runtimeconfig.json.local.example`.
+### Emulator Initial Setup & Configuration
 
-**Running the emulators**
+The emulators use a local configuration file instead of `functions:config:set`.
 
-!!! info "Important" - Do this before running the [client](#running-and-building-the-client). - Frankly does not currently operate with a named database; the name must be the default, literally the string `default`.
-
-To run the emulators locally, run the following _while_ in the `firebase/functions` directory:
+To setup, create the following file if it does not already exist:
 
 ```
-dart run build_runner build --output=build
-firebase emulators:start --only firestore,functions,auth,pubsub,database
+firebase/functions/.runtimeconfig.json
 ```
 
-You can also just run `npm run emulators`.
+A sample configuration file can be found at:
 
-We recommend using the emulators [import and export](https://firebase.google.com/docs/emulator-suite/connect_firestore#import_and_export_data) functionality to make development easier.
+```
+firebase/functions/.runtimeconfig.json.local.example
+```
 
-!!! question ""
-Please refer to the Cloud Functions Emulator [section](faq.md#cloud-functions-emulator) at [**❓Troubleshooting / FAQ**](faq.md) for common issues and resolutions!
+Copy the example and fill in any required values.
+
+---
+
+### Running the Emulators
+
+The easiest way to run the full emulator suite is with the project script.
+
+From the `firebase/functions` directory run:
+
+```
+npm run emulators
+```
+
+This script will:
+
+1. Build the Dart functions using `build_runner`
+2. Start the Firebase emulator suite for:
+   - Firestore
+   - Functions
+   - Auth
+   - Pub/Sub
+   - Realtime Database
+
+Alternatively you can run the underlying script directly:
+
+```
+# from within the firebase/functions directory:
+./emulators-start.sh
+```
+
+---
+
+### Restarting or Stopping the Emulators
+
+Firebase emulators spawn multiple background processes (including Java processes). If a previous run terminates unexpectedly, emulator ports may remain occupied and prevent startup.
+
+To safely stop all emulator processes run:
+
+```
+./emulators-stop.sh
+```
+
+Then restart them with:
+
+```
+./emulators-start.sh
+```
+
+Or with npm:
+
+```
+npm run emulators
+```
+
+---
+
+### Why the Stop Script Exists
+
+You may encounter errors like:
+
+```
+Port 8080 is already in use
+```
+
+This typically happens if emulator processes did not shut down cleanly.
+
+Do **not** attempt to fix this by killing Java processes manually, for example:
+
+```
+pkill -f java
+```
+
+That approach can terminate unrelated tools such as:
+
+- VS Code language servers
+- Gradle daemons
+- other development processes
+
+Instead, the project provides `emulators-stop.sh`, which performs a **deterministic shutdown**:
+
+1. Requests a graceful shutdown from the Firebase Emulator Hub (`localhost:4400`)
+2. Waits for emulator ports to actually close
+3. If necessary, terminates only processes listening on known emulator ports
+
+This ensures that unrelated processes are never affected.
+
+---
+
+### Inspecting Emulator Ports Manually
+
+If the emulators still fail to start, you can inspect which processes are using the emulator ports with:
+
+```
+lsof -nP -iTCP:4400 -iTCP:4000 -iTCP:8080 -iTCP:8085 -iTCP:9000 -iTCP:9099 -iTCP:5001 -sTCP:LISTEN
+```
+
+If anything appears in the output, run:
+
+```
+./emulators-stop.sh
+```
+
+to cleanly terminate those processes.
+
+---
+
+### Emulator Data Import / Export
+
+We recommend using the Firebase emulator import/export functionality to preserve development data between runs.
+
+Documentation:  
+<https://firebase.google.com/docs/emulator-suite/connect_firestore#import_and_export_data>
+
+This allows you to seed Firestore or other services with test data and reuse it across sessions.
+
+---
+
+### Important Notes
+
+- Run the emulators **before starting the Flutter client**.
+- Frankly currently expects the **default Firestore database name** (`default`).
+- Always run emulator commands from the `firebase/functions` directory.
 
 ### Optional: Setup Firebase Cloud Project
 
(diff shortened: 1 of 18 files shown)


firebase/functions/lib/utils/infra/firestore_utils.dart
```diff
@@ -26,7 +26,14 @@ class FirestoreUtils {
...
+    // Ensure `id` is populated from the document path. Models with
+    // `required String id` need this because Firestore doesn't store the
+    // document ID as a field — and `_$$_XFromJson` serializes `id: null`
+    // when writing, so putIfAbsent would be a no-op on read. Use ??= to
+    // overwrite null regardless of whether the key is present.
+    map['id'] ??= path.split('/').last;
+    return constructor(map);
   }
 
   Map<String, dynamic> fromFirestoreJson(Map<String, dynamic> json) {
```
katherineqian: Ditto for the other file changed in [d2bf7cb](https://github.com/berkmancenter/frankly/commit/d2bf7cbcc0273b227e65ea690d3cc524946800f2). It would be good to attach a PR containing this change to the issue for `JSNull` deserialization Sentry error (or create that issue).

docs/pages/dev.md
```diff
@@ -174,11 +174,44 @@ Copy the example and fill in any required values.
...
+6. Waits for emulators to become ready on their ports
+7. Launches the Flutter client pointing at the local emulators
+
+Steps 1–4 use stamp files in `.local/dev-stamps/` to skip unnecessary work.
+
+**Optional environment variables:**
+
+| Variable | Effect |
+|---|---|
+| `SKIP_DART_BUILD=1` | Skip the `build_runner` step in `emulators.sh` (set automatically by `run-dev.sh` since it handles the build itself) |
```
katherineqian: Already addressed my comment 😁

katherineqian (changes requested):
Awesome work @mikewillems! Requesting changes mostly to split out the null / doc ID fix from this PR since that's a separate bug which would be good to have tracked and addressed separately from these dev improvements. I especially appreciate your thorough documentation as always, super helpful for collaborators!

firebase/functions/emulators-stop.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
mikewillems: They were honestly really similar. I simplified things a bit in the latest version - see what you think!

katherineqian (approved):
Thanks @mikewillems, looks great and is so helpful for getting everything running properly in one command! Thanks for your hard work! 🥳

PR #352, opened 2026-03-20
Fix firestore JSNull deserialization errors

mikewillems:
## What This Fixes
Deserialization errors when Firestore documents have required String ID fields that end up as JSNull:
- Crashes with: `'JSNull is not a subtype of String'`
- Affects template lookups, community lookups, and other models with `'required String id'` fields
- Sentry error that users have been hitting

## Root Cause
When Firestore objects are deserialized via `getFirestoreObject()`, if a required ID field isn't explicitly set during initialization, it gets initialized as JSNull. The type system then fails when it expects a String.

## Solution
Initialize ID fields with the document's path-derived ID during deserialization, ensuring the field is never JSNull.

## Changes in the Code Base

Commits:
1. `d2bf7cbc`: Fix model lookup errors for Template, Community, and others
2. `f55c0d89`: Extend fix to JoinEvent participant (follow-up)

## Why Separate from Local DX PR?
- This is a product fix affecting the back end
- Local DX improvements are about developer workflow (emulators, scripts)
- Cleaner separation for review and issue tracking
- Can be merged independently

## Related
- Linked to Sentry error tracking JSNull deserialization issues
- Should reference/close any existing issue about model lookup failures 

None.

## Testing this PR
This fix is low-risk and narrowly scoped (just initializing ID fields from the document path during deserialization). Testing should focus on:

- **Smoke test**: Run the Firebase Functions test suite (`cd firebase/functions && npm test`) and confirm no regressions
- **Manual verification**: With local emulators running, trigger flows that load Templates, Communities, and JoinEvent participants - confirm no JSNull crashes in the console
- **Sentry monitoring**: After deploy, watch for recurrence of the `JSNull is not a subtype of String` error - it should stop appearing

No new unit tests are strictly required since the fix is defensive initialization, but if the test suite already covers `getFirestoreObject()` deserialization, those tests will implicitly validate this change.

### DIFF
--- a/firebase/functions/lib/utils/infra/firestore_utils.dart
+++ b/firebase/functions/lib/utils/infra/firestore_utils.dart
@@ -26,7 +26,14 @@ class FirestoreUtils {
     final ref = firestore.document(path);
     final snapshot =
         transaction == null ? await ref.get() : await transaction.get(ref);
-    return constructor(fromFirestoreJson(snapshot.data.toMap()));
+    final map = fromFirestoreJson(snapshot.data.toMap());
+    // Ensure `id` is populated from the document path. Models with
+    // `required String id` need this because Firestore doesn't store the
+    // document ID as a field — and `_$$_XFromJson` serializes `id: null`
+    // when writing, so putIfAbsent would be a no-op on read. Use ??= to
+    // overwrite null regardless of whether the key is present.
+    map['id'] ??= path.split('/').last;
+    return constructor(map);
   }
 
   Map<String, dynamic> fromFirestoreJson(Map<String, dynamic> json) {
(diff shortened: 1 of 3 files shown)


firebase/functions/lib/utils/infra/firestore_utils.dart
```diff
@@ -26,7 +26,14 @@ class FirestoreUtils {
     final ref = firestore.document(path);
     final snapshot =
         transaction == null ? await ref.get() : await transaction.get(ref);
-    return constructor(fromFirestoreJson(snapshot.data.toMap()));
+    final map = fromFirestoreJson(snapshot.data.toMap());
+    // Ensure `id` is populated from the document path. Models with
+    // `required String id` need this because Firestore doesn't store the
+    // document ID as a field — and `_$$_XFromJson` serializes `id: null`
+    // when writing, so putIfAbsent would be a no-op on read. Use ??= to
+    // overwrite null regardless of whether the key is present.
```
katherineqian: Thanks for the comment!

katherineqian (approved):
This fix looks good to me - sorry for the delay, and thanks for both fixing this and for separating it out from #337!

PR #354, opened 2026-03-20
Updated Fix for Hosted Recordings

### LINKED ISSUE #213: Unable to download long recordings
_Note: This was initially a bug report from a beta user and was tracked internally until we confirmed it was not a security issue. Now tracking here for visibility._

1. Record a full-length event
2. Attempt to download video recording

### Expected behavior

Download a zip file with recording

### Actual behavior

Either infinite loading wheel or (if it does complete) gives a user-facing error message and a CORS error upon inspection

### Additional details

Eric investigated enough to determine that the CORS error is not the root of the problem. This issue has to do with the exception thrown by the download recording function.

_No response_

### Operating system

_No response_

mikewillems:
## What is in this PR?

Recording downloads in the admin events tab worked in production but were fragile: it previously streamed recordings through a Firebase function to create a .zip archive via the `archiver` package, which failed for longer recordings due to ZIP size constraints and function response limits. This PR replaces the entire approach. The Firebase function now returns short-lived signed GCS URLs as JSON; the client downloads each recording part directly from GCS, with multiple files where needed to avoid the size constraint. 

In the admin events tab, UI is merged with the prior data improvements but expanded. The Recordings checkbox label shows a live status annotation, part count when ready, "Preparing" while MP4s are not yet available, and a failed note if the status check errors out. The Download button is enabled only when at least one ready item is selected. Status polling starts from page load so the state is typically resolved before the user opens the dialog. 

## Changes in the codebase

- `7c0fd524` - `firebase/functions/js/download-recordings.js`, `package.json`: Rewrote the `downloadRecording` Firebase function to authenticate with GCS, list MP4 files for the event, and return signed URLs as a JSON array instead of streaming a .zip. Changed error responses from `HttpsError` (callable-only format) to proper HTTP status codes. When no recordings are ready, returns `{ recordings: [] }` with HTTP 200 instead of a 404, so the client can distinguish "not ready" from "error."

- `35153911` + follow-ups - `client/lib/features/admin/presentation/views/data_tab.dart`: Added recording state polling to the admin events tab. On page load, each past `alwaysRecord` event triggers a status check against `downloadRecording`. State is tracked per event as: `null` = loading, `0` = not ready (retrying), `N` = N parts ready, `-1` = error (polling stopped). Retries every 5 seconds while count is 0, stops after 6 retries and surfaces an error state. Follow-up commits refine this: move kick-off to the stream listener (`cbe40b68`); fix status request to send JSON and suppress retry on hard HTTP errors (`0ba915e3`); add idToken null guard and retry-until-ready logic (`73175150`); fix async variable capture by storing locals before awaits (`8df95aba`); re-enable retry for errored recordings with a max retry cap (`e4723fb2`, `916f1411`). The initial UI surface (an inline recording widget) is superseded by `409a3349`; polling machinery stays intact though.

- `7d24bb32` + follow-ups - `client/lib/l10n/app_en.arb` and matching locale files: Initial l10n strings for recording status. Superseded by `37b76262`.

- `03d03004` - `firebase/functions/package-lock.json`: Lock file update removing `archiver` and its transitive dependencies.

- `135c54dc` - `firebase/functions/js/download-recordings.js`: Added `responseDisposition` to the `getSignedUrl` call so GCS embeds `Content-Disposition: attachment; filename="..."` in the signed URL, ensuring browsers download the file rather than open it inline.

- `9273320e` - `firebase/functions/lib/events/live_meetings/agora_api.dart`: Log the full Agora acquire response body and status code, and throw explicitly on non-2xx. Previously a failed acquire silently produced a null resource ID and let the start call fail with a confusing error downstream.

- `9000b65d` - `firebase/functions/emulators.sh`, `client/lib/config/environment.dart`: Two local dev fixes. In `emulators.sh`: before starting the emulators, check for `firebase/functions/service-account-key.json` and export it as `GOOGLE_APPLICATION_CREDENTIALS` if not already set; this enables `file.getSignedUrl()` in the local emulator without manual shell setup, and is a no-op if the file is absent. In `environment.dart`: changed `functionsUrlPrefix` from a compile-time const with a hardcoded `/dev/` project ID fallback to a static getter that constructs the emulator URL from `FIREBASE_PROJECT_ID` when `FUNCTIONS_URL_PREFIX` is not set; developers no longer need `FUNCTIONS_URL_PREFIX` in `.env`.

- `409a3349` - `client/lib/features/admin/presentation/views/data_tab.dart`: Restore the "Select Data" alert dialog. Remove inline recording widget and separate registrant-only download button. Replace both by a dialog with a Recordings checkbox and a Registrations checkbox, wired to `_downloadAllRecordings` and `downloadRegistrantList` respectively. Keep / use state polling from `35153911`.

- `5384b76c` - `client/lib/features/admin/presentation/views/data_tab.dart`: Wire recording status into the dialog. Each event's polling state is mirrored in a `ValueNotifier<int?>` (`_recordingNotifiers` map) so an open dialog receives live updates without a parent `setState` rebuild - Flutter's `showDialog` creates an independent route that does not re-render when the parent rebuilds. Inside the dialog, `StatefulBuilder` manages checkbox selection state and a nested `ValueListenableBuilder<int?>` subscribes to the notifier. The Recording checkbox label shows a status annotation via `_recordingAnnotation`; the checkbox is disabled while preparing (parts null or 0); the Download button is gated on `recordingReady` in addition to checkbox selection.

- `37b76262` - `client/lib/l10n/app_en.arb` and generated locale files: Adapted l10n strings for the dialog checkbox label context. Removed three ARB keys used only by the removed inline widget: `recordingNotAvailable`, `downloadRecordingParts`, `retryRecording`. Added four new keys for the status annotation: `recordingStatusChecking`, `recordingStatusPreparing`, `recordingStatusFailed`, `recordingStatusParts` (with `int count` placeholder). Updated the abstract `AppLocalizations` class and all three locale implementation files (EN, ES, ZH); ES and ZH stub the new annotation keys with English pending translation.

- `bc2bdbb6` - `client/lib/features/admin/presentation/views/data_tab.dart`, `app_en.arb`: Renamed dialog checkbox labels to shorter forms ("Recordings", "Registrations"). Changed Recording checkbox initialization: starts unchecked when recordings are still preparing, starts pre-checked only when recordings are already available at dialog open time. If recordings become available while the dialog is open, the `ValueListenableBuilder` fires an `addPostFrameCallback` to auto-check the Recording checkbox once; a `recordingAutoChecked` guard prevents re-checking if the user subsequently unchecks it.

### Minor follow-up commits

- `74d3a453` - "Preparing recording..." shown as a disabled button rather than plain text, for visual consistency with the download button.
- `0c27e4d3` - Consistent disabled button styling across states.

Three GCP/IAM changes are required for this PR to work in each environment, plus Firebase Functions config. These must be set manually per environment. For each environment, the following are needed for the principal listed in the table below:

- `Storage Object Creator` role on the recording bucket - covers HMAC writes by Agora
- `Storage Object Viewer` role on the recording bucket - covers `bucket.getFiles()` and reading files in the download function
- `Service Account Token Creator` role on the principal itself - covers signed URL generation
- Firebase Functions config: `agora.storage_bucket_name`, `agora.storage_access_key`, `agora.storage_secret_key`

**Local Dev**
Principal name: `frankly-dev-a410b@appspot.gserviceaccount.com`
Firebase/Functions project: `frankly-dev-a410b` (grant `serviceAccountTokenCreator` here)
Bucket project: `hls-bkc-asml-moot-dev` (grant storage roles and create HMAC key here)
Recording bucket name: `deliberations-agora-recordings-dev`
Status: Code-side setup is done (`9000b65d`). Remaining steps are manual external config: grant `storage.objectViewer` on the recording bucket to the local dev principal (for download); grant `storage.objectCreator` and create an HMAC key in `hls-bkc-asml-moot-dev` for Agora (for recording start); populate `firebase/functions/.runtimeconfig.json` with `agora.storage_access_key`, `agora.storage_secret_key`, `agora.storage_bucket_name`. `serviceAccountTokenCreator` is not required locally - see section 3 below.

**Staging**
Principal name: `gen-hls-bkc-7627@appspot.gserviceaccount.com`
Firebase/Functions project: `gen-hls-bkc-7627` (grant `serviceAccountTokenCreator` here)
Bucket project: `hls-bkc-asml-moot-dev` (grant storage roles and create HMAC key here)
Recording bucket name: `deliberations-agora-recordings-dev`
Status: DONE as of 2026-03-24.

**Prod**
Principal name: `asml-deliberations@appspot.gserviceaccount.com`
Firebase/Functions project: `asml-deliberations` (grant `serviceAccountTokenCreator` here)
Bucket project: `asml-deliberations` (same project - grant storage roles and create HMAC key here)
Recording bucket name: `deliberations-agora-recordings-prod`
Status: DONE as of 2026-03-24.

### Detailed Description of GCS Key Settings

**1. HMAC key for Agora to write recordings to GCS**

Agora uses HMAC (S3-compatible) credentials to write `.ts`, `.m3u8`, and `.mp4` files to the recordings bucket. The key configured in Functions config (`agora.storage_access_key` / `agora.storage_secret_key`) must exist and be active in GCS under the GCS project > Settings > Interoperability > Service Account HMAC Keys, and the associated service account must have `roles/storage.objectCreator` on the recording bucket.

This was absent for local dev (not yet set up) and was broken on staging before this PR (the configured key had been deleted over a year ago). Without this, recording start appears to succeed (Agora returns a valid `sid`) but no files ever appear in GCS.

**2. `roles/storage.objectViewer` on the recording bucket for the Cloud Function service account**

The `downloadRecording` function reads from the recording bucket using the Cloud Function's Application Default Credentials. When the bucket lives in a different GCP project from the function's host project (as is the case for staging, where functions run in `gen-hls-bkc-7627` but the bucket is in `hls-bkc-asml-moot-dev`), read access must be explicitly granted. Even when both are in the same project (prod), the role must still be assigned explicitly to the principal.

Without this, every call to `downloadRecording` returns HTTP 500 `{"error":"Failed to generate recording URLs"}` even when no recordings exist, because `bucket.getFiles()` throws before the empty-list check.

**3. `roles/iam.serviceAccountTokenCreator` for the Cloud Function service account**

The `downloadRecording` function calls `file.getSignedUrl()` to generate short-lived download URLs. On Cloud Functions Gen 1 with Application Default Credentials, the service account must have this role on itself in order to sign URLs. Grant it to the principal (from the table above) in that environment's IAM console.

This role is not exercised until there are MP4 files in the bucket. On staging and prod it has been granted preemptively. For local dev it is not required: when `GOOGLE_APPLICATION_CREDENTIALS` points to a service account key file, the Storage SDK signs URLs locally using the private key in the file rather than calling the IAM API.

## Testing this PR

1. Deploy the Firebase function to the target environment.
2. In the admin events tab, find a past event with `alwaysRecord` enabled that has completed recordings in GCS.
3. Click Download and verify the "Select Data" dialog opens with Recordings and Registrations checkboxes. Verify the Recordings label shows a status annotation (part count / preparing / failed), the checkbox is disabled while not ready, and the Download button is disabled when only a not-ready Recordings is selected.
4. Once recordings show a part count, select Recordings and click Download; verify each part arrives as a separate file (not a zip).
5. Select only Registrations and click Download; verify the registrant list downloads correctly.
6. For an event mid-processing, open the dialog and leave it open; verify the annotation updates live as parts become available and the Recordings checkbox auto-checks at that point.
7. Test a session over 2 hours long; all parts should download without size-limit failure.

## Additional Information

The `archiver` dependency is fully removed from `package.json`. It was the core of the previous zip-streaming implementation, not an incidental dependency.


client/lib/features/admin/presentation/views/data_tab.dart
dariusk, line 1: Straightforward, easy to read code. Love it

firebase/functions/js/download-recordings.js
```diff
@@ -1,83 +1,94 @@
...
-        const uid = decodedToken.uid
-
-        // Extract the eventPath from the request body
-        const { eventPath } = req.body
-        if (!eventPath) {
-            throw new functions.https.HttpsError('failed-precondition', 'eventPath not found')
-            return
-        }
+// Signed URLs expire after 15 minutes (milliseconds)
+const signedUrlExpiration = 15 * 60 * 1000
```
dariusk, line 12: So if the overall download process takes more than 15 minutes this will fail, right? Fwiw that seems fine given the file sizes and likely internet connections speeds involved but wanted to be clear.

firebase/functions/js/download-recordings.js
```diff
@@ -1,83 +1,94 @@
...
 
-        // Fetch the event object
-        const eventDoc = await firestore.doc(eventPath).get()
-        if (!eventDoc.exists) {
-            throw new functions.https.HttpsError('failed-precondition', 'event not found')
-        }
-        const event = { id: eventDoc.id, ...eventDoc.data() }
+// Expected path shape: community/{id}/templates/{id}/events/{id}
+const eventPathRegex = /^community\/[^\/]+\/templates\/[^\/]+\/events\/[^\/]+$/
```
dariusk, line 15: Is there a world where `eventPath` comes in with a trailing slash like `community/{id}/templates/{id}/events/{id}/`?
```suggestion
const eventPathRegex = /^community\/[^\/]+\/templates\/[^\/]+\/events\/[^\/]+\/?$/
```

firebase/functions/js/download-recordings.js
```diff
@@ -1,83 +1,94 @@
...
-        }
-        const membership = membershipDoc.data()
+const downloadRecording = functions.https.onRequest((req, res) => {
+    cors(req, res, async () => {
+        try {
+            const authToken = req.headers.authorization?.split('Bearer ')[1]
+            if (!authToken) {
+                res.status(401).json({ error: 'Unauthorized' })
+                return
+            }
```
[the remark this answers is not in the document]
dariusk: Lol this copilot suggestion seems like over-engineering - oh no, malformed authentication headers might be treated as unauthenticated! Tbh might be a useful suggestion for other kinds of headers but for auth it's pretty laughable.

firebase/functions/js/download-recordings.js
```diff
@@ -1,83 +1,94 @@
...
+            const membershipDoc = await firestore.doc(membershipPath).get()
+            if (!membershipDoc.exists) {
+                res.status(403).json({ error: 'membership not found' })
+                return
+            }
+            const membership = membershipDoc.data()
 
-            // Pipe the archive's output to the response
-            archive.pipe(res)
+            if (!['owner', 'admin'].includes(membership.status)) {
```
dariusk, line 54: I'd love for these to be enums and not strings but that's not for this PR...

firebase/functions/js/download-recordings.js
```diff
@@ -1,83 +1,94 @@
...
+            // from GCS without the function acting as an intermediary.
+            // responseDisposition embeds Content-Disposition: attachment in the
+            // signed URL itself, so browsers download rather than play inline.
+            const urls = await Promise.all(
+                mp4Files.map(async (file) => {
+                    const filename = file.name.split('/').pop()
+                    const [url] = await file.getSignedUrl({
+                        action: 'read',
+                        expires: Date.now() + signedUrlExpiration,
+                        responseDisposition: `attachment; filename="${filename}"`,
```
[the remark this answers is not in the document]
dariusk: This is false, at least the carriage return and newline part of the statement. From [Google's documentation](https://docs.cloud.google.com/storage/docs/objects#general_requirements):

> * Object names cannot contain [Carriage Return or Line Feed characters](https://en.wikipedia.org/wiki/Newline#Unicode).
>
>   * Note, backslashes sent through interfaces such as the gcloud CLI and Google Cloud console are escaped under the hood. For example, \n becomes \\\n. The Carriage Return and Line Feed restriction refers strictly to ANSI escape characters.

dariusk (approved):
Overall looks great! I cannot test this in my local environment but from a purely code review standpoint I left some comments (and some comments on a certain robot's comments).

client/lib/features/admin/presentation/views/data_tab.dart
mikewillems: Thanks!

firebase/functions/js/download-recordings.js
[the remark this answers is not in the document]
mikewillems: Lol yea I hadn't read this one yet - it's pretty funny.

firebase/functions/js/download-recordings.js
mikewillems: I actually have another PR that does exactly this kind of thing - I kind of just went hunting in the code base. I'll add this in there if it's not already 😆

firebase/functions/js/download-recordings.js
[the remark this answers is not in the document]
mikewillems: Thanks for the heads! Lol I might have fallen for that one.

firebase/functions/lib/events/live_meetings/agora_api.dart
```diff
@@ -82,6 +82,11 @@ class AgoraUtils {
       body: body,
     );
 
+    print('Acquire response (${result.statusCode}): ${result.body}');
+    if (result.statusCode < 200 || result.statusCode > 299) {
+      throw HttpsError(
+          HttpsError.internal, 'Acquire failed: ${result.body}', null);
```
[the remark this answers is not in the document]
mikewillems: I'll wrap this in a debug gate.

firebase/functions/js/download-recordings.js
mikewillems: Shouldn't be: the fullPath is `$collectionPath/$id`.

collectionPath is set from eventRef.parent.path - these are always generated by Firebase, and it won't produce a trailing slash there. The only risk would be if someone manually added it (like copied a longer path within an event and backspaced through the extension back to the event url but left the /).

firebase/functions/js/download-recordings.js
mikewillems: No, 15m is just the max time between signed URL generation and the download starting, but the URL isn't generated until the user clicks `Download`.

At that point the browser just needs to initiate each individual file download within those 15 minutes - which always happens within a few seconds since anchor.click() just schedules the download, it doesn't wait for the file transfer to complete.

PR #356, opened 2026-03-22
fix(accessibility): remove redundant Semantics label on SelectableLinkify

### LINKED ISSUE #340: Fix double semantic declaration for chatMessage.

In `firebase/functions/test/community/get_members_data_test.dart`, we declare the chatMessage semantic label twice in nested fashion on lines 151 and 194. That makes an awkward duplicate announcement on screen readers.

### Expected behavior

Screen reader only announces the chat message field once on entry / focus.

### Actual behavior

Screen reader announces the chat message field twice on entry / focus.

### Additional details

The inner Semantics around SelectableLinkify ([message_display.dart line 151](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)) should be removed. It labels only the raw text body as "chat message", which is both redundant (the outer already does this) and misleading (the text is just one part of the message, not the whole thing).

Any

### Operating system

Any

novavale:
## Summary

Fixes #340

The outer `Semantics` widget in `MessageDisplayState.build` already labels the entire chat message widget with `context.l10n.chatMessage`. The inner `Semantics` wrapping only the `SelectableLinkify` (raw message text) was redundant and misleading:

- **Redundant:** screen readers announced "chat message" twice on focus — once for the full widget, once for the text fragment alone.
- **Misleading:** the inner label described only the raw text as "chat message", which omits the sender name and timestamp that are part of the semantic whole.

## Change

Removed the inner `Semantics` wrapper from `message_display.dart` lines 154–168. `SelectableLinkify` now renders directly without an additional semantics node; the outer `Semantics` at the widget root still provides the single, correct announcement.

## Testing

No automated widget tests needed — this is a pure semantic tree reduction with no behavior change. Manual verification: focus a chat message with a screen reader and confirm it announces once, not twice.

### DIFF
--- a/client/lib/features/chat/presentation/widgets/message_display.dart
+++ b/client/lib/features/chat/presentation/widgets/message_display.dart
@@ -151,21 +151,17 @@ class MessageDisplayState extends State<MessageDisplay> {
                           // Note: There are highlighting issues due to the below
                           // https://github.com/Cretezy/flutter_linkify/issues/59
                           // https://github.com/Cretezy/flutter_linkify/issues/54
-                          Semantics(
-                            label: context.l10n.chatMessage,
-                            child: SelectableLinkify(
-                              text: widget.message.message ?? '',
-                              style:
-                                  context.theme.textTheme.bodyLarge!.copyWith(
-                                color: Theme.of(context).isDark
-                                    ? context.theme.colorScheme.onPrimary
-                                    : context.theme.colorScheme.primary,
-                              ),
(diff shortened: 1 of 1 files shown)


katherineqian (approved):
ty @novavale for addressing this!

katherineqian:
@novavale Note: going forward, please make your contributions against staging, as that is our working branch! I will need to revert and re-implement this change since there are significant merge conflicts in staging with your contribution.

PR #359, opened 2026-03-23
Make the CommunityProvider nullable

### LINKED ISSUE #358: CommunityProvider accessed without null check, causing console error
(lol the following was written manually by me, @dariusk, it just shows up as authored by Sentry because I did it from inside Sentry)

Recurring issue in prod where console errors are thrown during navigation transitions. When a user moves from a context where they have access to the "community nav" to one where they don't, sometimes `CommunityProvider` is briefly accessed during the transition in an attempt to rebuild the (non-existent) community nav. Changing line 162 in `lib/core/widgets/navbar/sidebar/sidebar.dart` should fix the issue.

Sentry Issue: [FRANKLY-EB](https://applied-social-media-lab-harva.sentry.io/issues/7295814375/?referrer=github_integration)

```
minified:q9: Provider<minified:n7> not found for minified:YU
  at Provider.of (../../../../../../../.pub-cache/hosted/pub.dev/provider-6.1.2/lib/src/provider.dart:343:7)
  at _SideBarState._buildMultipleCommunityNav.<anonymous function> (../../../../../../../.pub-cache/hosted/pub.dev/provider-6.1.2/lib/src/provider.dart:693:21)
  at _SideBarState._buildMultipleCommunityNav (../../../lib/core/widgets/navbar/sidebar/sidebar.dart:160:32)
  at A.bUE.prototype$2 (../../../lib/core/widgets/navbar/sidebar/sidebar.dart:151:20)
  at CustomStreamBuilder.build.<anonymous function> (../../../lib/core/widgets/custom_stream_builder.dart:94:16)
...
(5 additional frame(s) were not displayed)
```

dariusk:
Fixes #358, an error surfaced by Sentry that I don't believe caused any user-facing errors. Previously there was a boolean expression in a filter that would sometimes try to compare a property of `CommunityProvider` - but sometimes `CommunityProvider` is null, so accessing the property to evaluate the expression would throw an error.

This fix makes the boolean expression return `false` when there's no CommunityProvider available, instead of throwing an error.

## Testing this PR
There's not much to test - I did a smoke test and things were working fine, and you could access nav menus as expected. The real test is whether the bug shows up in production again after this fix is applied.

### DIFF
--- a/client/lib/core/widgets/navbar/sidebar/sidebar.dart
+++ b/client/lib/core/widgets/navbar/sidebar/sidebar.dart
@@ -159,7 +159,7 @@ class _SideBarState extends State<SideBar> {
         !isInitializedOnHome) {
       final currentCommunity = communities.firstWhere(
         (element) =>
-            context.watch<CommunityProvider>().communityId == element.id,
+            context.watch<CommunityProvider?>()?.communityId == element.id,
       );
 
       communities = [


katherineqian (approved):
Ideally we would fix the design causing this UI to depend on a sometimes-null value in the first place. However, I believe we'll refactor the sidebar at some point, and as a stopgap this makes sense.

PR #361, opened 2026-03-25
Updated analyzer to catch more issues.

mikewillems:
## What is in this PR?
Super atomic PR for making better use of our analyzer.

## Changes in the codebase
Only changing the `client/analysis_options.yaml` file.
- Add error for BuildContext across async gaps (previously unmarked).
- Upgrade unawated_futures (i.e. missing awaits) to error.
- Add warning for dead code.
- Upgrade prints to warning (prefer explicit logging).
- Add warning for undeclared dependencies.
- Warn for missing trailing commas. 

None

## Testing this PR
Nothing to test - but requesting dialog around the choices!

## Additional information
At the very least, the `use_build_context_synchronously: error` is a strong rec. This has been a thorn in our side and manifested in several discovered bugs.

### DIFF
--- a/client/analysis_options.yaml
+++ b/client/analysis_options.yaml
@@ -15,11 +15,20 @@ analyzer:
     - lib/**.g.dart
     - test/**.mocks.dart
   errors:
-    unnecessary_null_comparison: error
-    dead_null_aware_expression: warning
-    missing_required_param: error
+    # Bugs
+    use_build_context_synchronously: error
     unawaited_futures: error
+    unnecessary_null_comparison: error
+    missing_required_param: error
+
+    # Should be visible / fixed promptly
     unused_import: warning
+    unused_element: warning
+    dead_null_aware_expression: warning
+    avoid_print: warning
+    implementation_imports: warning
+    depend_on_referenced_packages: warning
+    require_trailing_commas: warning
 
     # invalid_annotation_target is recommended in the Freezed package documentation.
     invalid_annotation_target: ignore


client/analysis_options.yaml
```diff
@@ -15,11 +15,20 @@ analyzer:
...
-    unawaited_futures: error
+
+    # Should be visible / fixed promptly
     unused_import: warning
+    unused_element: warning
+    dead_null_aware_expression: warning
+    avoid_print: warning
+    implementation_imports: warning
+    depend_on_referenced_packages: warning
+    require_trailing_commas: warning
```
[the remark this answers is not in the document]
mikewillems: Lol, so unused variables and unreachable null checks aren't dead code now?

client/analysis_options.yaml
```diff
@@ -15,11 +15,20 @@ analyzer:
     - lib/**.g.dart
     - test/**.mocks.dart
   errors:
+    # Bugs
+    use_build_context_synchronously: error
```
katherineqian, line 19: I would count this as tech debt and wouldn't want this to be a blocker for all upcoming work - suggest for this to be marked as a warning instead. As you mentioned, we have a lot of these (just checked -- 58), and they should be fixed with a little care to avoid breaking things.

katherineqian (approved):
Just the one comment on the error-level lint - otherwise, happy to have these warnings added, thanks @mikewillems! 

I wonder if we can also silence all the hundreds of `.arb` warnings. Looking into it.

client/analysis_options.yaml
mikewillems: I hear that. My thought: even thought the right fix is the big fix that shall not be named, these still can cause crashes and should still be guarded now even as band-aids. Imo if it can crash the app, it's an error, even if it's tech debt. Plus, I already have a bunch fixed locally, and the band-aid is quick.

Also, the linter is only going to make a blocker out of files actively touched by a PR, not EVERYTHING. So really this means "we HAVE to fix these when we come across them instead of ignoring them", not "we can't proceed until we've fixed it all".

PR #365, opened 2026-03-28
Fix vote_to_kick.dart path typo in function-mapping.json

mikewillems:
## What is in this PR?

Currently, if we make changes to the VoteToKick firebase function, it won't trigger an auto redeploy of the functions upon push because of a typo in the function mapping listing; the path was `lib/event/` instead of `lib/events`. I noticed this when I was figuring out the downloadRecording function but wanted to separate it out so it's trivial to check.

## Changes in the codebase

Saving the file triggered a Prettier autoformat, so that's the first commit with no actual changes.
The second commit is just a 1 character fix of that typo:
`.github/workflows/function-mapping.json` line 55.

None.

## Testing this PR

Not needed.

## Additional information

None.


katherineqian (approved):
ty for typo fix!

PR #377, opened 2026-04-02
Remove blue borders around in-meeting admin dialogs + small UI fixes.

katherineqian:
## What is in this PR?

Currently, admin dialogs like "Breakouts" and "Jump to Room" still have a hardcoded bright blue border. This gets rid of those borders and fixes some other small color issues, removing hardcoded colors and fixing an invisible text color on an action button.

## Changes in the codebase

* 

*


mikewillems (approved):
Nice, super atomic! Always love to see styles abstracted!

Issue #371, opened 2026-04-01
Clarify backend file naming and organization for meeting / recording data


mikewillems:
Currently, we store back end data in a very confusing, difficult to navigate, and often impossible to search / sort structure of folders and files on GCP and Firestore. The pitch is to make this easier.

### Formats

**GCS recordings** (`agora_api.dart` `fileNamePrefix: [roomId]`, retrieved in `download-recordings.js`):

```
{bucket}/{eventId}/{agora-filename}.ts
{bucket}/{eventId}/{agora-filename}.m3u8
{bucket}/{eventId}/{agora-filename}.mp4
```

Bucket name from Functions config key `agora.storage_bucket_name`, e.g. `deliberations-agora-recordings-dev`.

**Firestore event data** (`FirestoreDatabase.communityCollectionName = 'community'` in `firestore_database.dart`):

```
community/{communityId}/templates/{templateId}/events/{eventId}
community/{communityId}/templates/{templateId}/events/{eventId}/event-participants/{userId}
community/{communityId}/templates/{templateId}/events/{eventId}/event-messages/{messageId}
community/{communityId}/templates/{templateId}/events/{eventId}/email-logs/{logId}
community/{communityId}/templates/{templateId}/events/{eventId}/live-meetings/{eventId}
community/{communityId}/templates/{templateId}/events/{eventId}/live-meetings/{eventId}/breakout-room-sessions/{sessionId}
community/{communityId}/templates/{templateId}/events/{eventId}/live-meetings/{eventId}/breakout-room-sessions/{sessionId}/breakout-rooms/{roomId}
community/{communityId}/templates/{templateId}/events/{eventId}/live-meetings/{eventId}/breakout-room-sessions/{sessionId}/breakout-rooms/{roomId}/live-meetings/{roomId}
```

**Client URL paths** (browser navigation, from `calendar_link_util.dart`, `email_templates.dart`):

```
/space/{communityId}/discuss/{templateId}/{eventId}
```

---

### Filtering and Sorting

**GCS recordings:**

- Can only be filtered by prefix (`bucket.getFiles({ prefix: '${eventId}/' })` in `download-recordings.js` line 61).
- No filtering by community, template, date, or event status without knowing the `eventId` upfront.
- Files within a prefix have Agora-generated names (no semantic meaning); not sortable by content type from path alone.

**Firestore event data:**

- Can query events scoped to a specific template/community via `FirestoreEventService.eventsCollection()`.
[...]

Issue #323, opened 2026-03-05
Live transcription


katherineqian:
#### Pitch

Every recorded session gets a per-room transcript automatically, with no manual steps. Agora RTT runs in parallel with Cloud Recording on the same channel - we start it when recording starts and stop it when recording stops. Transcript segments land in Firestore in real time as speech is recognized, also knocking out #49 (minus UI).

#### Implementation Approach

Scope: additions to `agora_api.dart`, new `AgoraRttWebhook` Cloud Function. Requires Phase 2. Enable RTT in Agora Console for the project (separate toggle from Cloud Recording).

1. Add `startTranscription({required String sessionId, required String roomId, required String webhookUrl})` to `agora_api.dart`:
   - Call `POST .../rtsc/speech-to-text/builderTokens` to get an RTT builder token (same app ID and credentials as recording).
   - Call `POST .../rtsc/speech-to-text/tasks` with the channel name, UID `457` (separate from recording UID `456`), the webhook URL, and language config (auto-detect or per-community).
   - Update the session document with `agoraRttTaskId`.
2. Add `stopTranscription({required String sessionId})`: read `agoraRttTaskId` from the session, call `DELETE .../rtsc/speech-to-text/tasks/{taskId}`.
3. Call `startTranscription()` immediately after `start` succeeds in `recordRoom()`; call `stopTranscription()` in `stopRoom()` before stopping Cloud Recording.
4. Add `AgoraRttWebhook` HTTP Cloud Function: resolve the active session by querying `roomId == cname && status == recording`, write each segment where `isFinal == true` to `transcript-segments`. Discard partial segments for now. Pass `${Environment.functionsUrlPrefix}/AgoraRttWebhook` into `startTranscription()` at call time.

PR #380, opened 2026-04-04
Implement Breakout Room Recordings

mikewillems:
## What is in this PR?

This PR adds both UI and functions changes to enable breakout room recordings with Agora (deterministically upon all methods of breakout room creation), map the recording files to appropriate and sensible locations in GCS, stream their availability status from Firestore in real time, convey that to the user via the UI - in their own language, and initiate all downloads at once.

### Caveats

This is still a relatively primitive version recordings. The main room recording still gets a blank section while participants are in breakouts, and breakout recordings seem to get blank sections after everyone returns to the main room. But a) this gets compressed away in terms of the file size so doesn't hurt there, b) this still makes all the actual video data available to clients, and their sensemaking platforms won't care unless they get charged by the minute, in which case they should honestly pre-process the video locally to speed it up 2x and halve their costs.

This is just how Agora gives the videos as .mp4 files, and doing it this way means sidestepping ingress/egress and processing costs in Firebase functions, as well as outsourcing QA and monitoring of the processing to Agora. For exceptionally long recordings (>4h or so), breakouts may run into [the same issue that we had for hosted meetings, #213](https://github.com/berkmancenter/frankly/issues/213), but we'll address that as needed.

## Changes in the codebase

Commits from oldest to newest:

- `**cdde8fc6**` Added `RecordingSession` and `TranscriptSegment` Freezed models in `data_models`. Top-level Firestore collection `recording-sessions` (not nested under events) simplifies admin queries and reduces collection overhead.
- `**80a6ca45**` Added `recordingSessionId` to `LiveMeeting` and `BreakoutRoom`. Check here first to avoid duplicate overwritten recordings on multiple joins / calls (makes it idempotent).
- `**21bffc6d**` Rewrote `agora_api.dart` recording lifecycle. Session doc is written before the Agora call (status: `starting`) so a crash mid-call leaves a recoverable record. GCS prefix changed from `[roomId]` to `[eventId, breakoutSessionId|'main', roomId, sessionId]` - prevents collision when the same room ID is reused across sessions, and scopes files per session for future per-session downloads. Part of [#371](<[https://github.com/berkmancenter/frankly/issues/371](https://github.com/berkmancenter/frankly/issues/371)>).
- `**c936e77d**` Idempotency guard in `live_meeting_utils.dart` and `getBreakoutRoomJoinInfo`. Previously, Agora silently swallowed duplicate `start` calls (error 4613); now session ID is written to the Firestore doc before Agora call, so the guard holds under concurrent joins. Also routes `eventId`, `breakoutSessionId`, `participantIds`, and `chatPath` through to the session doc so future transcript and chat export work has the right pointers from the start.
- `**58299e8f**` `event_ended.dart` now calls `stopRoom()` for the main room. Previously relied on Agora's own idle-timeout. Stop call is wrapped in try/catch so a failed Agora response won't block post-event email flow.
- `**bf4ad055**` `produce-sessions.js`: Firestore `onUpdate` trigger that fires on `status -> stopped`. Finds the Agora-produced `.mp4` under the session's GCS prefix and writes its path to `artifactPaths.complete_mp4`. Also exports any `transcript-segments` to a JSON file and writes its path to `artifactPaths.transcript_json` (no-op when RTT not implemented).
- `**f07272df**` Firestore composite index on `(eventId ASC, startedAt ASC)` for `recording-sessions`, and a security rule allowing authenticated reads. Without the index, the client-side `where eventId ==` query in DataTab would fail. Keep writes server-only.
- `**ce751b8e**` Registered `getSessionDownloadUrl` in the Dart CF entrypoint (`node/main.dart`). This was already implemented in JS but not connected; without this registration Firebase doesn't deploy the function.
- `**197fedca**` `event_ended.dart` now also iterates `breakout-room-sessions/{sessionId}/breakout-rooms` and calls `stopRoom()` for each room with a `recordingSessionId`. Each stop is individually try/caught. Also fixed a double-path bug: `liveMeetingPath` already ended with `/${event.id}`, so appending it again gave non-existant path.
- `**ee607671**` All room-construction sites in `assign_to_breakouts.dart` -- targetPerRoom loop, smart-match prematch, smart-match regular, waiting-room insertion, and empty fallback -- now set `record: true` when `alwaysRecord` is on, not just the smart-match prematch path (others were silently always off).
- `**003a43b4**` `eventEnded` call in `live_meeting_provider.dart` was gated behind three conditions none of which are routinely true in the emulator or for communities without the pre/post feature. Now also fires unconditionally when `alwaysRecord` is set. No change to the pre/post dialog gate.
- `**02536fd6**` and `**c98c2dc4**` `recordingSessionStream` added to `FirestoreLiveMeetingService`, and the in-meeting recording badge wired to it. Badge now hides when the session `status == failed` rather than staying on indefinitely. Falls back to always-visible before a session ID exists.
- `**beb1a468**` DataTab recording column replaced HTTP polling with a realtime Firestore stream over `recording-sessions where eventId ==`. Aggregate file count across all sessions (main + all breakout rooms) so the download link reflects the total. Status codes: `null` = loading, `-2` = at least one session recording, `0` = all stopped but no artifacts yet, `N>0` = N files ready, `-1` = stream error.
- `**d2618a19**` `useDatabaseEmulator()` throws if called twice on the same RTDB instance. Hot-reload cycles in dev were hitting this. Added a static guard flag.
- `**e5bc6651**` Moved breakout recording _start_ from join-time (`getBreakoutRoomJoinInfo`) to assignment-time (`_processAssignments`), just after rooms are written to Firestore. The server is the only writer at that point, which fixes the TOCTOU race where two concurrent joins both see `recordingSessionId == null` and each launch an Agora session on the same channel. The on-join guard is kept for selective-record events (non-`alwaysRecord`) where rooms are not pre-stamped. Fun note: we may have also been getting billed by Agora for the duplicate sessions.
- `**08a7c9df**` Localized the 3 new recording status strings (`recordingStatusInProgress`, `recordingStatusProcessing`, `recordingStatusFile`/`recordingStatusFiles`) across all four ARB files.

Firestore index and security rule changes (`firestore.indexes.json`, `firestore.rules`) are deployed alongside the functions but are technically still in the repo. The following manual steps are required after deploying the branch.

### Agora webhook secret

The `agoraRecordingWebhook` function verifies incoming Agora callbacks with HMAC-SHA1. The secret must be set in Firebase config per environment:

```bash
firebase functions:config:set agora.webhook_secret="<secret>" --project <project-id-will-need-to-do-again-for-prod>
firebase deploy --only functions --project <project-id>   # redeploy to pick up config
```

The secret value is copyable from Agora Console during NCS registration (next step). If not set, the webhook skips signature verification and logs a warning -- acceptable for dev, not for production.

### Agora Console: Message Notification Service (NCS) registration

Agora's NCS is a per-project webhook system configured in Agora Console. It POSTs JSON callbacks to an HTTPS endpoint when recording events occur. This is separate from the Agora REST API credentials.

Steps:

1. Log in to Agora Console (console.agora.io)
2. Projects -> find project -> Edit
3. All Features -> Notifications tab
4. Cloud Recording section -> configure:

- **Events**: select at minimum **41** (`recorder_leave` -- fires when the recorder bot exits the channel, i.e. recording stopped). Also added 40 (`recorder_started`) and 11 (`session_exit`) for observability.
- **Receiving Server Region**: North America (functions run in `us-central1`)
- **Receiving Server URL Endpoint**: `https://us-central1-<project-id>.cloudfunctions.net/agoraRecordingWebhook`

1. Copy the displayed secret -> use as `agora.webhook_secret` above
2. Click Check -- Agora sends synthetic callbacks (`cname: "test_webhook"`, `uid: "12121212"`) and expects 200 OK within 10s. The function handles this correctly (returns 200 for all event types other than 41).
3. If health check passes, click Apply Settings.

Note: eventType 41 is `recorder_leave` (recorder bot exits channel).

### Prod setup steps

1. Deploy functions + Firestore to staging
2. Agora stuff:
3. Set agora.webhook_secret in staging config
4. Register webhook URL in Agora Console NCS
5. Pass Agora health check and apply
6. Redeploy functions to pick up config

Steps 2 requires Agora Console access. Until completed, recording sessions that stop via Agora idle timeout (e.g. tab close without clicking Leave) will stay in `recording` state indefinitely and their MP4 files will exist in GCS but not appear in the UI.

## Testing this PR

Two manual tests cover nearly all recording paths. I already verified late-enable of recordings separately. Never hurts to test it more, but at least it's disjoint / not a confounding factor for these tests.

### Test 1: Simple main-to-breakout round-trip

1. Create a meeting with recording enabled. Start it.
2. Enter breakouts (1 room is fine).
3. Return to the main room.
4. End the event normally.
5. Check Admin > Event Data. Expect 1 main room file + 1 breakout file.

### Test 2: Multi-room with participant swaps

1. Create a meeting with recording enabled. Start it with 2 participants.
2. Enter breakouts with 2 rooms, 1 participant each.
3. Swap participants one at a time: move participant A into room B (so both are in room B), then move participant B out of room B into room A. Now each participant is in the other's original room.
4. Reunite into one breakout room. End breakouts.
5. One participant closes the tab. The other clicks "Leave".
6. Check Admin > Event Data. Expect files for the main room plus each breakout room that had a participant.

## Bug fixes (post-review)

These commits address bugs found during staging testing and PR review feedback.

- `**c8941c56**` `assign_to_breakouts.dart`: start breakout recordings when `liveMeeting.record` is set (the `?record=true` URL-param path), not only when `alwaysRecord` is on.
- `**caf5f54b**` `live_meeting_provider.dart`: restart main-room recording when participants return from breakouts. Previously only breakout rooms got re-recorded.
- `**b97ddca0**` `live_meeting_utils.dart`: restart breakout recording on re-entry after idle timeout. If Agora's idle timeout stopped a breakout recording and a participant rejoined, the recording did not resume.
- `**0d151584**` `agora_api.dart`, `assign_to_breakouts.dart`: structured logging of all breakout/video info at recording initiation for easier debugging of Agora interactions.
- `**d5eca120**` `agora-recording-webhook.js`: new Agora NCS webhook that transitions a recording session to `stopped` when Agora fires eventType 41 (`recorder_leave`). Prevents sessions from staying in `recording` state indefinitely if `eventEnded` misses them.
- `**b048acae**` `agora-recording-webhook.js`: fix payload field path (`payload.cname`/`payload.sid`, not `details.cname`) and use `req.rawBody` for HMAC verification instead of re-serialized JSON.
- `**6327339a**` `produce-sessions.js`: register all MP4 files found under the session GCS prefix as `complete_mp4_0`, `complete_mp4_1`, etc., instead of only the first.
- `**72e8ec0a**` ARB files: fill in Spanish, Simplified Chinese, and Traditional Chinese translations for all `recordingStatus*` keys (previously forgot to replace English placeholders).
- `**4ee5e4d3**` `data_tab.dart`: wrap snapshot data with `fromFirestoreJson()` so Firestore `Timestamp` objects are converted to `DateTime` before `RecordingSession.fromJson`. Without this, `startedAt`/`stoppedAt` silently deserialize as null.
- `**36d5c915**` `firestore.indexes.json`: move both `recording-sessions` composite indexes into the `indexes` array (they had been placed in `fieldOverrides` by mistake, causing deploy failures).
- `**9ef401ae**` `firestore.rules`: add TODO for scoping `recording-sessions` reads to community members (currently `hasAuth()`).
- `**5323245b**` `get_breakout_room_join_info.dart`: null guard on `currentBreakoutSession` -- throws `failedPrecondition` instead of interpolating `"null"` into the Firestore path.
- `**7a84b708**` `live_meeting_provider.dart`: `hadRecording` check now includes `liveMeeting?.record` so `eventEnded` fires for URL-param-triggered recordings too.
- `**d71a55ca**` `agora-recording-webhook.js`: add buffer length check before `timingSafeEqual` to avoid throwing on length mismatch from malformed signatures.
- `**f304131c**` `get_meeting_join_info.dart`, `live_meeting_utils.dart`: move `recordRoom()` call outside the Firestore transaction. Previously, if the tx was contested and retried, `recordRoom()` fired again and created duplicate Agora sessions. Now tx returns a `PendingRecording` struct and caller invokes Agora after the tx commits.


firebase/firestore/firestore.rules
```diff
@@ -487,5 +487,10 @@ service cloud.firestore {
             allow write: if false;
             allow read: if false;
         }
+
+        match /recording-sessions/{sessionId} {
+            allow read: if hasAuth();
```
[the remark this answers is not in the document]
mikewillems: We're not there yet, but I'll make a ticket.

firebase/functions/lib/events/live_meetings/breakouts/get_breakout_room_join_info.dart
```diff
@@ -73,11 +73,20 @@ class GetBreakoutRoomJoinInfo
...
 
+    final breakoutRoomPath =
+        '$liveMeetingPath/breakout-room-sessions/${liveMeeting.currentBreakoutSession?.breakoutRoomSessionId}/breakout-rooms/${request.breakoutRoomId}';
+
     final joinInfo = await liveMeetingUtils.getBreakoutRoomJoinInfo(
       communityId: event.communityId,
+      eventId: request.eventId,
+      breakoutSessionId:
+          liveMeeting.currentBreakoutSession?.breakoutRoomSessionId ?? '',
+      breakoutRoomPath: breakoutRoomPath,
```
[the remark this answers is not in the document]
mikewillems: Not really necessary, but good form, changed anyway.

firebase/functions/lib/events/live_meetings/live_meeting_utils.dart
```diff
@@ -62,6 +68,24 @@ class LiveMeetingUtils {
...
+    if (newSessionId != null) {
+      final chatPath =
+          '$liveMeetingCollectionPath/$meetingId/chats/community_chat/messages';
+      final participantIds = liveMeeting.participants
+          .map((p) => p.communityId)
+          .whereType<String>()
+          .toList();
+      await agoraUtils.recordRoom(
+        roomId: meetingId,
+        sessionId: newSessionId,
```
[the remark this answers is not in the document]
mikewillems: Important bug catch here. Fixed.

saihla:
@audreyt Alerting you to this big step!

Breakout recordings are working on this branch. Both we and Unify America have successfully tested it. We'll be making some adjustments and running additional tests before deploying in the coming weeks.

Thank you @mikewillems for all of your hard work on this!!!

audreyt:
Awesome work, thank you @mikewillems !

data_models/lib/recording/recording_session.dart
```diff
@@ -0,0 +1,78 @@
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:data_models/utils/firestore_utils.dart';
+
+part 'recording_session.freezed.dart';
+part 'recording_session.g.dart';
+
+enum RecordingSessionStatus {
+  starting,
```
katherineqian, line 8: Thanks for setting this up in such a clear way!

firebase/functions/lib/events/live_meetings/agora_api.dart
```diff
@@ -39,26 +43,140 @@ class AgoraUtils {
...
+      );
+      await sessionRef.updateData(UpdateData.fromMap({
+        'agoraSid': sid,
+        RecordingSession.kFieldStatus: RecordingSessionStatus.recording.name,
+      }));
     } catch (e) {
-      print("Error in starting recording for room $roomId");
-      print(e);
+      print('Agora start failed for room $roomId: $e');
+      await sessionRef.updateData(UpdateData.fromMap(
```
katherineqian, line 121: Nice status tracking!

firebase/functions/lib/events/live_meetings/live_meeting_utils.dart
```diff
@@ -74,22 +98,44 @@ class LiveMeetingUtils {
...
+    required String? existingRecordingSessionId,
+    required List<String> participantIds,
   }) async {
     final token =
         agoraUtils.createToken(uid: uidToInt(userId), roomId: meetingId);
-    if (record) {
-      await agoraUtils.recordRoom(roomId: meetingId);
+
+    if (record && existingRecordingSessionId == null) {
+      final newSessionId = _uuid.v4();
```
katherineqian: Any reason you went for UUID doc IDs instead of the standard Firestore doc ID? If no reason in particular, I'd lean towards using Firestore doc IDs to mirror our existing document format / for readability of the paths.

firebase/functions/js/produce-sessions.js
```diff
@@ -0,0 +1,88 @@
...
+                await change.after.ref.update({
+                    'artifactPaths.complete_mp4': mp4Path,
+                })
+                console.log(`Registered MP4 ${mp4Path} for session ${sessionId}`)
+            }
+        } catch (err) {
+            console.error(`Error registering MP4 for session ${sessionId}:`, err)
+        }
+
+        // --- Export transcript ---
```
katherineqian: It would be helpful to remove this section from this PR to keep the changes within the current scope - we can re-introduce when it's used / linked to actually generating transcript segments for #323. Even if it's a no-op now, we can't properly test this logic until the data is actually flowing. Let’s keep it out for now so we can verify it correctly in the next PR.

firebase/firestore/firestore.rules
katherineqian: Can we restrict this to mods + admins for the community? You can use some of the existing conditions in this rules doc. I see this is a TODO, but it should be achievable using existing patterns in `firestore.rules`.

firebase/functions/lib/events/live_meetings/breakouts/assign_to_breakouts.dart
```diff
@@ -278,6 +280,7 @@ class AssignToBreakouts {
     profile('total matches: ${prematches.length + matches.length}');
 
     final breakoutMatchIdsToRecord = event.breakoutMatchIdsToRecord.toSet();
+    final alwaysRecordSmart = event.eventSettings?.alwaysRecord ?? false;
```
katherineqian: suggestion: `alwaysRecordSmart` -> `alwaysRecord` - I wasn't sure what the `smart` part referred to when I read it below (I understand that it's being used within smart match, but the actual variable is just using the event settings rather than some kind of smart recording toggle).

client/lib/features/events/features/live_meeting/data/providers/live_meeting_provider.dart
```diff
@@ -706,16 +706,22 @@ class LiveMeetingProvider with ChangeNotifier {
       final timeNow = clockService.now();
       final event = eventProvider.event;
 
-      if (prePostEnabled && postEventCardData != null) {
```
katherineqian, line 709: I wonder why this condition was in place? Any insight?

client/lib/features/events/features/live_meeting/features/video/presentation/views/video_flutter_meeting.dart
```diff
@@ -146,42 +147,69 @@ class _VideoFlutterMeetingState extends State<VideoFlutterMeeting> {
...
+                  _buildRecordingBadge(context),
               ],
             ),
           ),
         ],
       ),
     );
   }
+
+  Widget _buildRecordingBadge(BuildContext context) {
```
katherineqian: Can you create a stateless widget for this + UI additions going forward, rather than a widget function? Widget functions are [not recommended](https://stackoverflow.com/questions/53234825/what-is-the-difference-between-functions-and-classes-to-create-reusable-widgets/53234826#53234826). Widget functions are used extremely frequently throughout this code base, and one example very relevant to us is that this contributes to `BuildContext` being used improperly (an error we see very often -- I highly recommend taking a look at the linked SO answer for more detail). I also suspect that the performance of the app could be greatly improved by removing widget functions.

My approach has been to remove widget functions when my work touches them, and for this reason I would not suggest adding additional widget functions to our code!

client/lib/features/events/features/live_meeting/features/video/presentation/views/video_flutter_meeting.dart
```diff
@@ -146,42 +147,69 @@ class _VideoFlutterMeetingState extends State<VideoFlutterMeeting> {
...
+                ),
+              ),
+              SizedBox(width: 26),
+            ],
+          ),
+        ),
+      );
+    }
+
+    // No session ID yet — show the badge unconditionally
```
katherineqian: Thanks for leaving these comments!

client/lib/app.dart
```diff
@@ -53,6 +53,8 @@ Future<void> reportError(dynamic error, dynamic stackTrace) async {
   }
 }
 
+var _databaseEmulatorConfigured = false;
```
katherineqian, line 56: Nit: Would be preferable to submit this commit as a separate PR in the future. No need to change it this time, but separate PRs are safer in case we experience issues with the main PR / faster for instant approve on small things like this.

firebase/functions/lib/events/live_meetings/breakouts/assign_to_breakouts.dart
```diff
@@ -677,6 +679,36 @@ class AssignToBreakouts {
       firstAgendaItemId: firstAgendaItemId,
     );
 
+    // Start recordings immediately after room assignment so there is exactly
+    // one writer and no risk of concurrent joins racing on the same room.
```
katherineqian, line 682: Great call!

firebase.json
```diff
@@ -95,7 +95,17 @@
     {
       "source": "firebase/functions",
       "codebase": "default",
-      "ignore": ["**", "node_modules", ".git"]
+      "ignore": [
```
katherineqian, line 98: Can you separate out this change so we can track it? As in my other comments, as an infra/config change, it'd be preferable for this to be visible at a top level instead of looped into "Breakout Room Recordings" in case we want to modify again in the future.

Edit: Okay, I was confused because I thought you had already addressed this issue. I checked the merge conflicts and now understand the timeline. Can we either take the version of `firebase.json` from staging or make a separate PR for this change?

client/lib/features/events/features/live_meeting/data/providers/live_meeting_provider.dart
```diff
@@ -787,6 +789,31 @@ class LiveMeetingProvider with ChangeNotifier {
     notifyListeners();
   }
 
+  Future<void> _restartMainRoomRecordingIfNeeded() async {
```
katherineqian, line 793: Appreciate the attention to detail here!

firebase/functions/lib/events/live_meetings/agora_api.dart
```diff
@@ -76,9 +76,14 @@ class AgoraUtils {
       ),
     ));
 
+    print(
+        'recording_start: sessionId=$sessionId roomId=$roomId eventId=$eventId roomType=${roomType.name} breakoutSessionId=$breakoutSessionId');
+
     String resourceId;
     try {
       resourceId = await _acquireResourceId(roomId: roomId);
+      print(
```
katherineqian, line 85: ty for adding logs 🙌

firebase/functions/js/agora-recording-webhook.js
```diff
@@ -0,0 +1,92 @@
...
+const admin = require('firebase-admin')
+const crypto = require('crypto')
+
+const firestore = admin.firestore()
+
+// Agora cloud recording event types
+// https://docs.agora.io/en/cloud-recording/develop/event-types
+const EVENT_TYPE_RECORDER_EXIT = 45
+
+// Agora signs each notification with HMAC-SHA1 over the raw body using the
```
katherineqian, line 12: Should we add this to the docs for setting up Agora?

firebase/functions/lib/events/live_meetings/get_meeting_join_info.dart
```diff
@@ -25,7 +25,7 @@ class GetMeetingJoinInfo extends OnCallMethod<GetMeetingJoinInfoRequest> {
     GetMeetingJoinInfoRequest request,
     CallableContext context,
   ) async {
-    return firestore.runTransaction((transaction) async {
+    final result = await firestore.runTransaction((transaction) async {
```
katherineqian, line 28: > If the tx was contested and retries (likely with multiple people **joining breakouts** at once),  recordRoom() tried again and sometimes started duplicate fAgora recordings (each with fresh UUID generated inside the callback).

To confirm, the recording is now starting after assignments are generated, not joining, right? Asking because this commit message comes after the one that moves the recording call. 

BTW, the order of commits listed in the PR description differs significantly from the timestamped/ordered list of commits here on GitHub - do you know what may be causing this discrepancy?

katherineqian (changes requested):
Thanks for your high-impact and well-documented work here @mikewillems! No significant structural changes, mostly a few requests to separate out code for maintainability + to refactor a widget.

client/lib/features/events/features/live_meeting/features/video/presentation/views/video_flutter_meeting.dart
mikewillems: Good call - thanks for the care here!

firebase/functions/lib/events/live_meetings/agora_api.dart
mikewillems: Thanks!

firebase/functions/lib/events/live_meetings/live_meeting_utils.dart
mikewillems: Mainly we just need an id before write; it requires a Firestore doc ref in order to create the id then which complicates things slightly and creates an async constraint. Good point that it is how we do it elsewhere, but it's also a non-obvious place to couple us to Firestore should we want to platform-generalize later. But it's pretty trivial and I agree with your leaning.

firebase/functions/js/produce-sessions.js
mikewillems: Yep, that's cleaner.

firebase/firestore/firestore.rules
mikewillems: Absolutely - I left it permissive for testing and forgot to go back and put that in.

mikewillems: Great catch!

firebase/functions/lib/events/live_meetings/breakouts/assign_to_breakouts.dart
mikewillems: I don't need to have a separate aliasing variable for this - this was back when I was using this in more sites for the transcript functionality; I'll just inline it instead.

firebase/functions/js/produce-sessions.js
mikewillems: (This was cherry-picked from an earlier branch that was covering transcripts too).

client/lib/features/events/features/live_meeting/data/providers/live_meeting_provider.dart
mikewillems: The conditions predate our entire Git repo, but looks like were there to gate "prePost" behavior that was alternative to the admin actually filling out postEventCard. So I imagine it roughly looked like answering the questions:
- is post event enabled? 
- has the user overridden = customized it?

Despite the name (which made me probably unwisely work from this var), `eventEnded` previously triggered only the post-event email flow, and the prePostEnabled && postEventCardData != null guard was there I'm assuming because that cloud function call only made sense in the context of post-event emails. The logic was: "only send the event-ended signal if there's actually a post-event experience to trigger." When I added recording, I initially inherited the same gate even though it has nothing to do with emails or post-event cards, so I moved it down to only apply where relevant.

firebase.json
mikewillems: This diff is misleading, this file already actually matches staging. So yes :)

firebase/functions/lib/events/live_meetings/get_meeting_join_info.dart
mikewillems: Yes confirmed, and the labeling is all a mess, my apologies. I thought I fixed most of it. I pulled most of this from my prior branch that was using the old chunk-based processing and integrating transcription into that. I can go back and correct it if you think that's worth it.

firebase/functions/lib/events/live_meetings/live_meeting_utils.dart
katherineqian: Firestore allows you to synchronously get a doc ref using collection.doc() which gives you an auto-generated document ID, so if this is the main reason I would suggest using generated document IDs. Using a custom UUID format for these documents wouldn't be super high impact for the larger goal of being platform agnostic IMO so I think the tradeoff for keeping our db consistent and readable is worth it in this case, unless there are other considerations you had in mind.

firebase/functions/lib/events/live_meetings/get_meeting_join_info.dart
katherineqian: That's fine, just wanted to clarify, thanks!

docs/pages/dev.md
```diff
@@ -419,6 +422,10 @@ agora.storage_secret_key="<YOUR_VALUE_HERE>"
   - `storage_secret_key`: From the generated key, paste the **Secret**.
 - **In the codebase**
   - In `client/.env`, change `FUNCTIONS_URL_PREFIX` to reflect your Google Cloud Run Functions prefix.
+- **Agora recording webhook**
+  - Deploy the `AgoraRecordingWebhook` Cloud Function and note its URL.
+  - In the Agora console, go to **Notifications** (under Developer Toolkit / Developer Resources). Add a new rule targeting **Cloud Recording** events, set the callback URL to the deployed function URL, and choose a secret string.
+  - `webhook_secret`: Set this to the secret string you chose above so the function can verify Agora's HMAC-SHA1 signature on each notification. If omitted, signature verification is skipped (acceptable for local development, not for production).
```
katherineqian, line 428: 🙌

katherineqian (changes requested):
Just the one remaining comment to discuss here. Thanks for addressing all these items so quickly!

docs/pages/dev.md
mikewillems: 🙌 
Btw, if you're wondering why my emoji reactions seem off - it's because Github only gives me these 8 options 😆  
image

firebase/functions/lib/events/live_meetings/live_meeting_utils.dart
mikewillems: Agreed - made the change for both instances where I was using uuid instead of Firestore native ids.

mikewillems:
Thanks @katherineqian for incredible work and attention to detail in the review process!

katherineqian (approved):
Thank you @mikewillems for this effort!

PR #384, opened 2026-04-08
Adding a link to the event editing page from event data

dariusk:
This adds a piece of text and a button to link to the event editing page so a user in the current release can find chat and poll data. This will be removed once we unify the download functionality into the event data tab.

image


katherineqian (approved):
LGTM - thanks for adding the i18n too!

PR #386, opened 2026-04-10
Fixing link copy

dariusk:
Updating copy as seen here

image


mwmarino:
@dariusk implementation of button appears to have no border/fill styling - we need something to further indicate visually this is interactive. Let's discuss Monday.

PR #388, opened 2026-04-13
Issue 310: Implement Flutter Web accessibility improvement workaround

### LINKED ISSUE #310: Implement Flutter Web accessiblity improvement workaround
Currently, our Flutter Web app suffers from relatively poor achievement of accessibility standards: see #296 (minimal screen reader support) for an example of an end user reporting issues with the accessibility of our site. While we have already made efforts to improve accessibility from the original state of the product by enabling Semantics in Flutter throughout key parts of the product, there is still much room for improvement in this area as demonstrated by #296. 

Additionally, inability to support pinch-to-zoom is a known issue with Flutter Web that also impacts our product, and causes us to fail to be able to meet known accessibility guidelines like WCAG (see [SC 2.5.1](https://w3.org/TR/WCAG22#pointer-gestures): Pointer Gestures (Level A)). 

An accessibility workaround has been proposed in this thread regarding [pinch-to-zoom](https://github.com/flutter/flutter/issues/97305) support in Flutter Web. In the best case, we would be able to build upon this workaround to meet accessibility guidelines. 

This issue represents the work of implementing the accessibility workaround proposed in that thread, as well as evaluating further opportunities to expand on the workaround for future improvements to accessibility.

jlemanski1:
## What is in this PR?
The goal of this PR is to improve the Lighthouse accessibility score. The primary issue is caused by Flutter's default `FullPageEmbeddingStrategy` which injects a viewport meta tag with `user-scalable=no` and `maximum-scale=1.0`. These cause WCAG2 compliance failures from automated scans like Lighthouse.

To work around this issue without waiting for the [upstream engine fix](https://github.com/flutter/flutter/pull/182047) to land in a stable Flutter release, this PR uses the `CustomElementEmbeddingStrategy` by rendering the app into a full screen container. This bypasses the Flutter engine's viewport injection, allowing a more accessible viewport tag to be defined.

## Changes in the codebase

* `client/web/index.html`
    * Added an accessible viewport meta tag
    * Added an element for mountint the Flutter app (`#flutter-container`)
    * Added CSS styling for body and `#flutter-container` to handle sizing, scroll suppression, and touch event handling
    * Removed the old `main.dart.js` and `platformVersion` script tags which are replaced with `flutter_bootstrap.js`

* `client/web/flutter_bootstrap.js`
  * Added `flutter_bootstrap.js` which replaces the old loading approach with Flutter's modern bootstrap API ([docs](FullPageEmbeddingStrategy))
  * Initializes the Flutter engine using the `#flutter-container` div as the host element, which uses Flutter's `CustomElementEmbeddingStrategy` as opposed to `FullPageEmbeddingStrategy` which overrides any defined viewport tags.
    * Its only responsible for bootstrapping the runtime, onece `appRunner.runApp()` is called, the Flutter engine hands off to whatever Dart entry point was compiled into the app.
      * This is why the `main.dart.js` script was removed from the index.html
      * The entry point is already in all of the workflows that I could see as well as in the `run_dev.sh` bash script, meaning no changes are required for how the app is built or run.

* None

## Testing this PR

This workaround makes no functional changes to the web app, but there is potential for unexpected behaviour between the browser and flutter web (mostly due to the architecture of Flutter Web). The CSS added to `index.html` ensures that Flutter receives touch and drag events before the browser intercepts them. That being said, browser/style compatibility is something to be aware of, so the majority of regression testing should focus on mobile devices. The following actions should be verified throughout the app:

* Verify that drag to scroll works
* Verify that pinch to zoom works where expected
( Verify that text fields focus correctly and the layout is recalculated/adapts for the presence of the onscreen keyboard.

The simplest way to test changes from your local on real mobile devices is to run the app with the `web-server` device and use a service like NGrok to temporary expose the web service for easy access from a phone.
```
# Run the app as a web server
flutter run -d web-server --web-renderer html -t lib/dev_emulators_main.dart --dart-define-from-file=.env --web-port=8888

# Expose the port using ngrok
ngrok http 8888
```

## Additional information

* This change gets Frankly to a 100 Lighthouse Accessibility score but does not necessarily guarantee any improvements for Screen Readers. The `Semantics` widget should still be used for translating Flutter widgets into ARIA roles.
  * Accessibility mode can be turned on explicitly with the following addition to `client/lib/main.dart`:
```dart
import 'package:flutter/semantics.dart';

void main() {
  runApp(const MyApp());
  if (kIsWeb) {
    SemanticsBinding.instance.ensureSemantics();
  }
}
```
This is turned off by default for performance reasons. Additional info can be found on the [Flutter Web accessibility page in the docs](https://docs.flutter.dev/ui/accessibility/web-accessibility#opt-in-web-accessibility)

* Loading states / progress indicators for the initial app load can be implemented in `flutter_bootstrap.js` ([example](https://docs.flutter.dev/platform-integration/web/initialization#example-display-a-progress-indicator))
* Registering ServiceWorkers is now handled in `flutter_boostrap.js` rather than script tags in the index.html


client/web/flutter_bootstrap.js
```diff
@@ -0,0 +1,21 @@
...
+
+_flutter.loader.load({
+  onEntrypointLoaded: async function(engineInitializer) {
+    /*
+     * Initialize the app runner inside the "flutter-container" element to 
+     * use the Flutter Engine's CustomElementEmbeddingStrategy. This is a 
+     * workaround to improve accessibility ratings by bypassing the viewport 
+     * attributes generated by the FullPageEmbeddingStrategy.
+     * 
+     * Note: #flutter-container requires `touch-action: none` in CC to allow
```
jlemanski1: - [x] Typo: CC -> CSS

client/web/index.html
```diff
@@ -91,19 +108,15 @@
   -->
   <script src="./agora/iris-web-rtc.js"></script>
 
-
-  <script type="application/javascript">
-    // Keep this in sync with the main.dart.js version below
-    window.platformVersion = '__VERSION__';
```
johnnycrich, line 97: We actually need to retain this for our cache busting. Thanks!

client/web/index.html
```diff
@@ -91,19 +108,15 @@
...
-
   <script>
     if (window.location.hostname == 'localhost') {
       // needed to make the auth emulator work correctly
       window.indexedDB.deleteDatabase('firebaseLocalStorageDb');
     }
   </script>
+  
+  <div id="flutter-container" role="main"></div>
+  <script src="flutter_bootstrap.js" async></script>
```
johnnycrich: We will need to append `?version=__VERSION__` to the filename. See above. :)

johnnycrich (changes requested):
Hello @jlemanski1, thanks so much for submitting this! Just a couple tiny changes needed. My schedule is booked for this week, but first thing Monday I will be testing this fix!

johnnycrich:
@jlemanski1 This fix looks good on my end -- please review the small requested changes and let me know if you have any questions, thanks!

jlemanski1:
> @jlemanski1 This fix looks good on my end -- please review the small requested changes and let me know if you have any questions, thanks!

Thanks! I'll get that change in later today

johnnycrich (approved):
After further testing, I found that this change does not allow for pinch-to-zoom, however it _does_ improve benchmarks. Thanks for the contribution!

PR #392, opened 2026-04-13
Template renaming- third attempt

karenunify:
### What is in this PR?
This PR renames the template blueprint component from "Guide" to "Template"

### Changes in the codebase
It was discovered in the analytics code that the term "guide" was sometimes used interchangeably with "template".

All references to the event template as the blueprint for the guide, but not the guide itself, should now be referred to as the template.


karenunify:
jk, this branch has too many merge conflicts

PR #409, opened 2026-04-29
Remove duplicative login/signup options and nav

### LINKED ISSUE #401: Remove Redundant Login Button
Currently the login option is in two places. We'd like it in one place for simplicity.

johnnycrich:
## What is in this PR?
This simplifies the sign-up/login experience and navigation logic.
- The sign-up and login buttons no longer trigger a modal popup and do not appear on the app's root (`/`) page. 
- The sidebar no longer shows sign-in options for signed-out users; instead, it shows an empty space, and the bottom sidebar buttons are styled differently depending on sign-in state. 

## Changes in the codebase
* The `SignInWidget` now navigates to the `HomeLocation` with a parameter to control whether the login dialog is shown by default, instead of opening a dialog directly. The sign-up button no longer triggers the login dialog either. 
* The `HomePage` and `HomePageSignInSection` components accept a `showLoginByDefault` parameter, which determines whether to show the login or sign-up UI by default. This parameter is passed through the routing logic. 
* Consolidated `Sidebar` layout methods; close button is now an `IconButton` with improved accessibility, styling.
* The navigation bar's profile/login widget now only shows login/sign-up options when not on the home page, using the new `showLoginSignUp` parameter.
* Unused imports and methods related to the old sign-in dialog flow are removed.

## Testing this PR

- [ ] After applying this change, note that sign-up/login buttons no longer trigger a modal, and do not appear on the app's root page when signed out. 
- [ ] Also note that the sidebar no longer contains a sign-up or login form when signed out, and appears as depicted:
Screenshot 2026-04-29 at 14 48 54

### DIFF
--- a/client/lib/core/routing/locations.dart
+++ b/client/lib/core/routing/locations.dart
@@ -83,12 +83,16 @@ class HomeLocation extends BeamLocation<BeamState> {
   @override
   List<String> get pathPatterns => ['/home'];
 
+  final bool showLoginByDefault;
+
+  HomeLocation({this.showLoginByDefault = false});
+
   @override
   List<BeamPage> buildPages(BuildContext context, BeamState state) => [
         _buildBeamPage(
           key: ValueKey('home'),
           title: context.l10n.appNameHome(Environment.appName),
-          child: HomePage(),
+          child: HomePage(showLoginByDefault: showLoginByDefault),
         ),
       ];
 }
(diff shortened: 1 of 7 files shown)


mwmarino:
Screenshot LGTM

dariusk (approved):
This all looks good to me, and importantly, it works nicely in testing. 

I would simply ignore the copilot suggestions as they don't make a ton of sense (or rather, are very nitpicky, for example "This will now render nothing for signed-out users on non-home pages where the bottom nav is used." is like... fine, since afaik there is no situation where the latter actually appears in a user flow)

PR #413, opened 2026-05-01
Move event data downloads to Event Data tab of Community Admin page + add a loading state to download button.

### LINKED ISSUE #311: Add visible loading state/feedback after requesting to download data
Currently, when selecting the download data option for polls, suggestions, registration data, chats, etc. there is no visible feedback/loading state after the user has pressed the button. This makes it look like nothing is happening for a few seconds until the action finishes and shows a snackbar/a file starts downloading.

**Proposed solution**
As a simple fix, I propose adding snackbars after users select "download data" buttons which simply say "Loading...". 

**Steps to complete**
- [ ] Find all "download data" buttons
- [ ] Add consistent "Loading..." snackbar on tap for all buttons

### LINKED ISSUE #345: Move event data export buttons to the new community admin UI "Data" tab.
#223 and #233 were developed in parallel, but #223 moved data exports to a new tab within Community Admin, while #233 added new data export buttons. This ticket is for moving those new data export buttons (for poll data, suggestions, chats, member registration) to the new "Data" tab within the Community Admin experience.

katherineqian:
## What is in this PR?

As in the title, this PR addresses #345. This removes the temporary redirect link placeholder we added in #384, and moves all event data downloads to one place as we had wanted to -- the Event Data tab of community admin. As related work, I also addressed #311 by adding a loading state to the new comprehensive event data download dialog.

Now, downloads are centralized in one location, which makes it a lot more convenient for organizers to access. The loading state for download buttons just improves the UX tremendously, as previously it seemed like the UI was just broken (when in fact the user had initiated a backend request, but the UI had not given any visual feedback). 

## Changes in the codebase

* In order to accommodate the extra complexity of the Event Data Download dialog, refactored it into its own `StatefulWidget` vs. a `StatefulBuilder`. 
* Refactored the Data Tab to accommodate using this new widget. 
* Added ability to download chats and poll/suggestions from the new dialog.
* Added functions to request chat + poll/suggest data when the dialog is opened, so user can preview whether this data actually exists for the event. It also shows how much data there is for each type, following the convention established for recordings ("n parts"). 
* Relaxes condition for showing "Data Download" button. 
    * Whereas it was hidden before if recording/registration data didn't exist, given that we are adding additional export types and may continue to add more, for performance reasons I started to move away from this approach. We can just show the button in most cases and users can load/select their desired data once the dialog is open, as opposed to loading all data for all events just so we can determine whether to show the button.
* Added a loading state for the download dialog.
* Removed all event data downloads from the event page info section.

* None.

## Testing this PR

* Go to the Community Admin > Event Data page. Attempt to:
    - [ ] Download recordings, registration data, chat data, and poll/suggestion data for an event that this exists for. Observe that all downloads should WAI and not be blank, and the data should be correctly formatted and accurate. 
    - [ ] Attempt to download chat data, and poll/suggestion data when there are no entries. Observe that it should not work. 
- [ ] Go to the Event Info page and attempt to download event data there. Observe that the page should be updated to no longer contain event data download links. 

## Additional information

N/A


client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
dariusk, line 1: Very glad to have this functionality moved into its own widget!

client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
```diff
@@ -0,0 +1,500 @@
...
+    // Reconstruct the response format for the CSV generator
+    final response = GetMeetingChatsSuggestionsDataResponse(
+      chatsSuggestionsList: chatData,
+    );
+
+    await provider.generateChatDataCsv(
+      response: response,
+      eventId: event.id,
+      breakoutRooms: breakoutRooms,
+    );
```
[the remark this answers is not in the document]
dariusk: This seems like a reasonable suggestion, just to happen right after we are done with `provider` here?

client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
```diff
@@ -0,0 +1,500 @@
...
+      event,
+      communityProvider: widget.communityProvider,
+    );
+    provider.initialize();
+    await provider.generatePollsSuggestionsDataCsv(
+      suggestionData: suggestionData,
+      pollData: pollData,
+      eventId: event.id,
+      breakoutRooms: breakoutRooms,
+    );
```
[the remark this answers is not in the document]
dariusk: Same as above, seems reasonable to me.

client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
```diff
@@ -0,0 +1,489 @@
...
+  }
+
+  Future<void> _loadEventData() async {
+    try {
+      // Load chat and suggestions data
+      final chatSuggestionRequest = GetMeetingChatsSuggestionsDataRequest(
+        eventPath: widget.event.fullPath,
+      );
+      final chatSuggestionResponse = await cloudFunctions.callFunction(
+        'GetMeetingChatSuggestionData',
```
dariusk, line 339: Random thing but I notice we have `GetMeetingChatSuggestionData` and `GetMeetingChatsSuggestionsDataRequest` (singular vs plural). Not a huge thing but probably something to rename in the future unless I'm missing something semantically different between the two.

client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
```diff
@@ -236,6 +236,8 @@ class _EventInfoState extends State<EventInfo> {
       existingTitles,
     );
 
+    if (!mounted) return;
+
```
dariusk, line 240: Just wondering what this is doing here in particular. Was there some kind of failure state we were running into?

client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
```diff
@@ -268,111 +270,14 @@ class _EventInfoState extends State<EventInfo> {
       onConfirm: (context) async {
         await alertOnError(context, () async {
           await _presenter.refreshEvent();
+          if (!context.mounted) return;
```
dariusk, line 273: Similar to above, and also is there a reason why this is on this particular line (after `refreshEvent` but before `Navigator.pop`)?

dariusk (approved):
Overall looks good! Had some questions about a few pieces, and some of the copilot suggestions seem okay (disposing the `EventProvider` in a few places), but I'm also no expert on Flutter lifecycles...

client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
katherineqian: This is true and a kind of confusing discrepancy. I was going to rename some of these functions but am a little wary of breaking things so I didn't think it was worth the hassle at this moment. If it's ok with you, I'm leaning towards leaving as-is for now.

client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
katherineqian: Just a check that prevents `context` from being called after it's already been disposed after an async gap. It's a pretty common error we get. It suppresses an alert flagged by a lint rule called `use_build_context_synchronously` (https://dart.dev/tools/linter-rules/use_build_context_synchronously).

client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
katherineqian: An async gap is created by `refreshEvent`, so guards against popping the `context` if the `context` was unmounted during `refreshEvent`!

client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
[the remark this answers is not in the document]
katherineqian: Done!

katherineqian:
Thanks for the prompt and detailed review, @dariusk!

Issue #48, opened 2025-02-11
Ability to end an event


saihla:
Currently, events end when everyone leaves. It would be useful for a host to be able to end an event, at which point participants would go through the exit flow and be placed back on the event page. Critical for this are defining the After-end behaviors.

# Working Acceptance Criteria

- [ ]  Internal machinery for event end / cleanup - after which attendees see a notice that the Event has ended and a link to go back to the org page.
- [ ]  For hostless events, enable automatic event end by time (option during setup?)
- [ ]  Hosts (specifically, users with Moderator, Admin or Owner role) can click a button to end an event immediately, regardless of event type (hosted, hostless, livestream).
- [ ]  Hosts receive a confirmation popup to prevent them from accidentally ending the event early
- [ ]  When event ends, participants go through the exit flow in the same way they would if they had clicked “Leave”

# Stretch Goals or futher possible tickets
- [ ]  Enable breakout room and event keep-alive until (1m after) last participant leaves or disconnects.
- [ ]  Add organizer option to choose whether to keep event alive until last participant leaves, immediately end and initiate exit flow, or keep alive for up to ___ minutes.
- [ ]  Put Slack-style hold music in single-person breakout rooms 😆 

- Unclear event endings (no host-initiated termination)

### Main Problem: Organizers Cannot End Events

**Symptoms**:

- Events continue indefinitely until last person leaves
- No apparent host control to forcefully end events
- Recordings unclear (do they stop? continue indefinitely?)
- Unclear post-event flow

**Root Causes** (verified in client code):

1. **No `endEvent()` method exists** in `live_meeting_provider.dart`
   - `endBreakoutRooms()` exists (line ~878) for ending breakout sessions
   - No corresponding `endEvent()` or similar host-initiated termination
2. **Existing `eventEnded` cloud function serves different purpose**
   - Called when individual participants leave (lines 681-730 in provider)
   - Appears oriented toward post-event email processing
   - Does not forcefully end events for all participants
3. **No client observer for event end state**
   - No evidence of listeners watching for "event ended" signal from server
   - Post-event flow only triggered by individual's voluntary leave action

**Assumptions**:

- At event end, we should probably have a grace period time where users can keep exchanging contact info and getting final thoughts down before the inputs stop.
- After that, instead of closing down immediately like Zoom and taking away all the data / chat, I think we should just show a read-only chat, but this would be a #design choice when we cross that bridge.
[...]

PR #223, opened 2025-08-07
Overhauled Community Admin UI

johnnycrich:
## What is in this PR?

As per #186, this change set re-implements the community admin interface to use M3 components, add new functionality where relevant.

## Changes in the codebase

* Usr `TabBarView`, built-in `TabController` rather than `CustomTabBarView` as the encompassing containment for all admin sections.
* Remove permissions enum `MembershipStatus.attendee` and update all others descriptions.
* Migrate usage of `CupertinoSwitch` to Material `Switch`.
* Add prefixIcon to `CustomTextField`.
* Where possible, utilized `Flex` widget with conditional direction based on device type in order to simplify layout code.
* Refactored overview tab to replace now deprecated settings panel `CreateCommunityDialog`.
* Migrated members tab to utilize Material `PaginatedDataTable`, allow for real time filtering of members.
* Created new `MemberDataUtils` class, new `DataTab` class, remove deprecated `EventsTab`.
* Overhauled `ChooseColorSection` to use a tab controller for preset/custom colors selection, new grid format for presets. Implemented [flutter_colorpicker](https://pub.dev/packages/flutter_colorpicker) for custom color selection.
* Consolidated community image service code within `CloudFunctionsCommunityService`.
* Removed code related to community banner image.
* Overhauled `SettingsTab` to divide toggles into sections, use updated Material components. 
* Removed "Optional" text from optional text fields, added a "*Required" helper for Name
* Community URL auto-updates when updating the community name, displays the updated community URL name if edited
* Use proper AlertDialog for error utils alerts as well as admin overview form issues.
* Added email, website, social link fields to `Community` model and overview form.
* Added dozens of new localization strings.
* For firebase test workflow, added install of JDK 21 so a firebase tools runtime can be installed.
* Near-countless other small changes.

Fixes #264 

## Testing this PR

- [x] Click on gear icon in navbar and see that you're taken to new overview section.
- [x] For all tabs, see that both the desktop and mobile layouts functions as you would expect, matches our [mockups](https://www.figma.com/design/PUxnXIYcSMqGOe5PKdZQhD/Frankly-Design-File?node-id=1006-5412&t=mA9hsjKcU4g1X6CJ-0).
- [x] Try editing various aspects of your community within this section, including removing and replacing image, updating URL, etc.
- [x] Give your community a new theme in the presets and see that it is reflected after saving and returning to landing page.
- [x] Try to give your community an invalid contrast for custom color theme, and see an error displayed to that effect.
[...]

PR #417, opened 2026-05-02
Automatic and Manual Event Ending

mikewillems:
# Automatic and Manual Event Ending

This implements [#48](https://github.com/berkmancenter/frankly/issues/48).

## What is in this PR?

Establish and test a consistent mechanism for ending an event manually and at a particular time. Enable the host to close all meeting rooms manually with a button or automatically at end time (with an optional grace period before closing all rooms). Add and test the UI for this functionality, and limit the grace period to an integer number of minutes between 0 (default) and 120.

### Event Timing Locations

Generally this has been cleaned up and clarified. After this PR, event timing is organized as follows:

Datum | How it's stored / accessed
-- | --
Scheduled start | Event.scheduledTime
Planned duration | Event.durationInMinutes
Scheduled end | Computed: scheduledTime + durationInMinutes
Actual end | LiveMeeting.events.last.timestamp where .event == finishMeeting

### General Approach

The core concept is a `meetingEndedAt` timestamp on the `LiveMeeting` document. Setting this field triggers two things: (1) the client detects the change via its live meeting stream and auto-leaves, and (2) server-side cleanup runs (stop recordings, send "event ended" email).

Two server paths can set `meetingEndedAt`:

- **Manual:** The host clicks "End Meeting" in the admin panel, which calls the `endMeetingForAll` Cloud Function. Authorization checks confirm the caller is the event creator or a community mod/admin/owner.
- **Automatic:** On first join, `GetMeetingJoinInfo` schedules a Cloud Task for `scheduledTime + durationInMinutes + gracePeriod`. When the task fires, `ScheduledEndMeeting` sets `meetingEndedAt` via the same shared core logic.

Both paths use a Firestore transaction for the idempotency guard: if `meetingEndedAt` is already set, the operation is a no-op. This prevents duplicate emails and recording stops even if the manual and scheduled paths race.

On the client, the event settings drawer exposes the `autoEndMeeting` toggle and a grace period input field. The grace period is validated client-side (0-120 integer, with debounced auto-correction) and clamped server-side as a defense-in-depth measure.

### Non-obvious choices

- **Grace period input uses a managed `TextEditingController` rather than `TextFormField` with `initialValue`.** The naive approach (using a `ValueKey` keyed to the model value) causes Flutter to destroy and recreate the `TextFormField` widget on every keystroke, because `onChanged` -> `updateSettingValue` -> `setState` changes the key. This also unfocuses the field. The managed controller approach avoids this by keeping the controller alive across rebuilds, with a `_gracePeriodEditing` flag to prevent `updateView` from overwriting the text while the user is typing.

- **Auto-end task scheduling is best-effort (try/catch).** A Cloud Tasks failure during `GetMeetingJoinInfo` would otherwise block the user from joining the meeting. The task is non-critical -- the host can always end manually.

- **`ScheduledEndMeeting` validates the `X-CloudTasks-TaskName` header.** Cloud Functions strips this header from external HTTP requests, so its presence confirms the call originated from a Cloud Tasks queue. This prevents unauthorized callers from ending meetings via the raw HTTP endpoint.

- **Recording stops and email delivery run in parallel.** These are independent operations. Running them sequentially risked approaching the Cloud Function timeout for events with many breakout rooms and gives inconsistent user experience (not that it really matters in this case but polish is nice).

- **Dialog dismiss is treated as cancel in the host leave prompt.** On my first pass, dismissing the "End for all?" dialog (tapping outside) forced the result to `false` and silently left the meeting. Now it cancels the leave action entirely, requiring an explicit choice.

### Changes by Commit

1. `7a3b4310` -- Add `meetingEndedAt` field to `LiveMeeting`, create `endMeetingForAll` CF with auth checks, extract recording stop logic into `stop_all_event_recordings.dart`
2. `34cc52d1` -- Add unit tests for `EndMeetingForAll` CF
3. `b190f08a` -- Client reacts to `meetingEndedAt` changes (auto-leave), wire up `endMeetingForAll` call, add "End Meeting" button in admin panel
4. `547212af` -- Add `ScheduledEndMeeting` CF (Cloud Tasks HTTP handler), schedule auto-end on first join in `GetMeetingJoinInfo`
5. `c4c34713` -- Add unit tests for `ScheduledEndMeeting` CF
6. `29107097` -- Comment `_postEventEmailThresholdInMinutes` explaining its purpose
7. `887a9e1c` -- Display end time alongside start time in event UI (carousel, event cards, event info)
8. `4fc78c4d` -- Add widget tests for `VerticalTimeAndDateIndicator` end time display
9. `5992828e` -- Reject join attempts on ended meetings (server-side check in `GetMeetingJoinInfo`)
10. `aa1fde93` -- Add test for join rejection on ended meetings
11. `b2ef50a8` -- Add host leave prompt ("End for all?" dialog before leaving)
12. `62bf5e14` -- Add `autoEndMeeting` and `autoEndGracePeriodMinutes` fields to `EventSettings`, add toggle + grace period input to event settings drawer, wire into task scheduling
13. `b42c2bd5` -- Fix settings type error by filtering non-bool values from the dev settings bool display
14. `4e27021f` -- Wrap the `meetingEndedAt` read-then-write in a Firestore transaction for atomic idempotency (both `EndMeetingForAll` and `ScheduledEndMeeting`)
15. `9a3b0c7b` -- Remove dead `displayName` resolution code and unused imports from `GetMeetingJoinInfo`
16. `2a3df808` -- Pass event from the transaction result via `MeetingJoinResult` instead of re-reading it from Firestore
17. `97ee210c` -- Wrap auto-end task scheduling in try/catch so join is not blocked by Cloud Tasks failures
18. `09e4eb52` -- Localize hard-coded strings in leave prompt and event settings (ARB files)
19. `0ed29a9b` -- Validate `X-CloudTasks-TaskName` header in `ScheduledEndMeeting` to prevent unauthorized external calls
20. `b42c8d3f` -- Add email verification to scheduled end meeting test
21. `295e7db5` -- Clamp grace period to 0-120 minutes server-side in `GetMeetingJoinInfo`
22. `b56e9a01` -- Add grace period input validation and auto-correction (managed controller, debounce, red error styling)
23. `e428a3fe` -- Extract shared end meeting core logic (transaction, recording stop, email) into `end_meeting_core.dart`, called by both CFs
24. `5f89aa07` -- Treat dialog dismiss as cancel in host leave prompt
25. `aafef468` -- Parallelize breakout room recording stops with `Future.wait`
26. `72f7f3e4` -- Run recording stops in parallel with email delivery in end meeting core
27. `0938e944` -- Use `permissionDenied` instead of `failedPrecondition` for auth errors (better error categorization in logs)
28. `a978fdcb` -- Short-circuit auth check for event creator in `endMeetingForAll` (skip membership lookup when the caller is the host)
29. `d3b0dcd9` -- Guard against double-tap on leave with `_leavingInProgress` flag
30. `3b6f104d` -- Add widget tests for grace period input validation and auto-correction
31. `cb49d913` -- Add comments explaining the validation scheme

No infrastructure or environment changes. Cloud Tasks is already enabled in the project. The `ScheduledEndMeeting` function is registered in `main.dart` and deployed alongside other Cloud Functions.

## Testing this PR

**Automated tests:**

- `EndMeetingForAll` CF unit tests (`end_meeting_for_all_test.dart`) -- covers auth, idempotency, recording stop, email send
- `ScheduledEndMeeting` CF unit tests (`scheduled_end_meeting_test.dart`) -- covers execution, idempotency, email verify
- `GetMeetingJoinInfo` test for join rejection on ended meetings
- Widget tests for end time display in `VerticalTimeAndDateIndicator`
- Widget tests for grace period input (initial value, focus retention, error styling, debounce auto-correction, focus loss correction, error clearing)

Run CF tests:

```bash
cd firebase/functions && CLOUD_RUNTIME_CONFIG=./test/test_config.json \
  firebase emulators:exec --only firestore --project frankly-dev-a410b \
  'dart test --concurrency=1'
```

Run client tests:

```bash
cd client && .fvm/flutter_sdk/bin/flutter test --platform chrome
```

**Manual testing checklist:**

- Toggle "Auto-end meeting" on in event settings, set a grace period, save, and verify the meeting auto-ends at the scheduled time
- Click "End Meeting" in the admin panel during a live meeting and verify all participants are ejected
- Verify the "End for all?" prompt appears when the host clicks Leave, and that dismissing it cancels the leave action
- Type invalid values (negative, >120, non-numeric, decimal) in the grace period field and verify red styling + auto-correction
- Join an ended meeting and verify the join is rejected

## Additional information

The `meetingEndedAt` field is write-once. Once set, there is no mechanism to "restart" a meeting. This is intentional -- ending is a final action confirmed by a dialog.

The grace period field is clamped to 0-120 on both client and server. The server-side clamp in `GetMeetingJoinInfo` is an extra defense (and good hygeine in case of unvalidated input attacks + is coupled with authentication); the client-side validation is the primary UX.

The `ScheduledEndMeeting` Cloud Function is an HTTP handler (not a callable) because Cloud Tasks dispatches via HTTP POST. The `X-CloudTasks-TaskName` header check is the auth mechanism -- Cloud Functions infrastructure strips this header from external requests.


client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
```diff
@@ -524,27 +526,96 @@ class _EventInfoState extends State<EventInfo> {
...
+                    unawaited(launch(snapshot.outlookCalendarLink));
+                    break;
+                  case CalendarMenuSelection.ical:
+                    _downloadICSfile(snapshot.icsLink);
+                }
+              },
+              type: ActionButtonType.outline,
+              color: context.theme.colorScheme.surfaceContainerLowest,
+              icon: Icon(
+                CupertinoIcons.calendar_badge_plus,
```
dariusk: I thought we were moving away from `CupertinoIcons`? This was my understanding from working on #223

client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
```diff
@@ -562,6 +633,23 @@ class _EventInfoState extends State<EventInfo> {
     );
   }
 
+  Widget _buildMessageParticipantsButton() {
+    return ActionButton(
+      onPressed: widget.onMessagePressed,
+      type: ActionButtonType.outline,
+      color: context.theme.colorScheme.surfaceContainerLowest,
+      icon: Icon(
+        CupertinoIcons.paperplane,
```
dariusk: See above on `CupertinoIcons`

client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
```diff
@@ -562,6 +633,23 @@ class _EventInfoState extends State<EventInfo> {
...
+        size: 24,
+        color: context.theme.colorScheme.onSurfaceVariant,
+      ),
+      text: 'Message participants',
```
dariusk: Should be a localized string

client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
```diff
@@ -793,6 +864,36 @@ class _EventInfoState extends State<EventInfo> {
...
+                  builder: (context) {
+                    final scheduledTime = eventProvider.event.scheduledTime;
+                    if (scheduledTime == null) return SizedBox.shrink();
+                    final date = DateFormat('EEE, MMM d').format(scheduledTime);
+                    final startFmt = DateFormat('h:mma').format(scheduledTime);
+                    final start = startFmt.substring(0, startFmt.length - 1).toLowerCase();
+                    final endDateTime = scheduledTime.add(
+                      Duration(minutes: eventProvider.event.durationInMinutes),
+                    );
+                    final endFmt = DateFormat('h:mma').format(endDateTime);
```
dariusk: Could we possibly reuse some of the time formatting code you add in [time_indicator.dart](https://github.com/berkmancenter/frankly/pull/417/changes#diff-75e83e996499317be389ba8c1f903479c5a486496ad793dea65e3dfd18b4a2bdR25-R33) here?

client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
```diff
@@ -840,39 +941,25 @@ class _EventInfoState extends State<EventInfo> {
...
-                            .canCancelEvent &&
-                        _event.status != EventStatus.canceled) ...[
-                      Align(
-                        alignment: Alignment.centerRight,
-                        child: _buildCancelEventButton(),
-                      ),
-                    ],
-                  ],
-                ),
+                if (canCancelParticipation || _canEditEvent) ...[                  SizedBox(height: 4),
```
dariusk: Strange whitespace                needs fixing!

firebase/functions/lib/events/live_meetings/end_meeting_for_all.dart
```diff
@@ -0,0 +1,87 @@
...
+      final snap = await transaction.get(liveMeetingRef);
+      final liveMeeting = LiveMeeting.fromJson(
+        firestoreUtils.fromFirestoreJson(snap.data.toMap()),
+      );
+      if (liveMeeting.meetingEndedAt != null) {
+        return false;
+      }
+      transaction.update(
+        liveMeetingRef,
+        UpdateData.fromMap({
```
dariusk, line 69: I think right around here we need some hooks for the Matomo API. Basically, wherever we have the call to set the meeting end time (or if it's in two places put it in both places), we should put code close to or identical to what's here:

https://github.com/berkmancenter/frankly/blob/2ac7c138a37d8e548d214354c39773e7457df8d6/client/lib/features/events/features/live_meeting/features/meeting_agenda/data/providers/meeting_agenda_provider.dart#L401-L410

and remove the linked code from `meeting_agenda_provider.dart`. The linked code is from an attempt to capture a "Complete Event" event for Matomo analytics. It logs 1) that the Frankly event ended, and 2) how many seconds long the event was. However, the linked code was written back before we really had a concept of "meeting ended" so it's really scattershot and only triggers sometimes. Deleting that code and dropping it in here (with a little math to calculate `duration`) would be ideal.

If we get this working right we'll be able to see in Matomo a list of anonymized meeting IDs alongside duration of the meeting, like in the following screenshot but with more than just the top entry working correctly

Image

firebase/functions/lib/events/live_meetings/get_meeting_join_info.dart
```diff
@@ -82,6 +79,34 @@ class GetMeetingJoinInfo extends OnCallMethod<GetMeetingJoinInfoRequest> {
       );
     }
 
+    // On first join, schedule automatic meeting end if the event has
+    // autoEndMeeting enabled, a scheduled time, and a duration.
+    if (result.isFirstJoin) {
```
dariusk, line 84: This whole conditional branch pleases me to look at

firebase/functions/lib/events/live_meetings/end_meeting_for_all.dart
dariusk: Oh yeah, probably also the hook needs to go in `firebase/functions/lib/events/live_meetings/scheduled_end_meeting.dart`

dariusk (dismissed):
This generally looks great! Love to see the debounces and clamps and all that bread and butter logic flow stuff.

I have more specific comments inline, particularly with some instructions on putting in Matomo hooks for analytics.

🫡 Been a pleasure serving with you!!

client/lib/features/events/features/event_page/presentation/widgets/event_info.dart
katherineqian: Yes, this is true. Please use `Icons` (Material Icons) for consistency!

firebase/functions/lib/events/live_meetings/get_meeting_join_info.dart
mikewillems: @dariusk your reviews are so much fun to read now that I'm finally getting back around to it. Thanks so much for the attention to detail! Merging soon :)

mikewillems:
@katherineqian I fixed all your and Darius's requests; only the last 7 commits since `f79b0b5` (merges included) are new to address them, changing these files:

- `date_utils.dart` -- added `formatTimeShort()` as separate utility
- `time_indicator.dart` -- switched to using `formatTimeShort()`
- `event_info.dart` -- formatting fixes, localization, icon replacements, time formatting refactor
- `meeting_agenda_provider.dart` -- removed client-side analytics call that only fired on full agenda completion -- now handled server-side in `endMeetingCore` for all end paths.
- `end_meeting_core.dart` -- added `_logMeetingEndAnalytics` (logs event ID + duration to Segment/Matomo) and `callerUid` param to attribute the end to whoever triggered it
- `end_meeting_for_all.dart` -- pass caller's auth UID to `endMeetingCore` so analytics can distinguish host-initiated ends from scheduled ones
- `client/lib/l10n/app_en/es/zh/zh_Hant_TW.arb` -- merged alphabetization; added `featured`, `autoEndMeeting`, `endMeeting`, `endMeetingConfirmation`, `gracePeriodMinutes`, `leaveWithoutEnding`
`firebase/functions/lib/admin/payments/analytics_util.dart` -- null protect _secretKey; skip analytics without error when Segment config is absent (fixes test crashes from our new endMeetingCore analytics call)

Many thanks for your thorough review!

firebase/functions/lib/events/live_meetings/get_meeting_join_info.dart
johnnycrich, line 84: @mikewillems I am unsure if this condition will ever return `true` at this point in the code path. If you look at [CreateEventDialogModel](https://github.com/berkmancenter/frankly/blob/6ef4443a1d9dab0396ef3fae575917ef6d02a5dd/client/lib/features/events/features/create_event/data/providers/create_event_dialog_model.dart#L227), the firestore event service is used to create an event doc immediately when the event is initially created.  `result.isFirstJoin` here is an alias of `!liveMeetingSnapshot.exists` (see `LiveMeetingUtils`), but because this code path is reached on meeting _join_, I don't see an instance where the document snapshot does not yet exist. I just traced this part of the code locally, and the event scheduler for ending the event later does not fire, even though I enabled auto-end. I also verified in the firestore emulator that the event document indeed pre-exists the event starting, e.g., the first participant join. Can you verify for me that when first joining an event, you indeed see this scheduler run? I am wondering, if using the value returned by `isNullOrEmpty(liveMeeting.meetingId)` might work here, since a meeting ID is not added to the doc until this function is first called within a meeting.

johnnycrich (changes requested):

firebase/functions/lib/events/live_meetings/get_meeting_join_info.dart
mikewillems: You are absolutely right - I modified to capture before the meeting start. Should fix that!

PR #420, opened 2026-05-05
fix: Migrate Community onboarding step stored enum values to new values (guide -> template). 

katherineqian:
## What is in this PR?

This is a fix intended to resolve the frontend client breaking after migrating enum values for Communities' stored `onboardingStep` values. 

## Changes in the codebase

* [Add manual JSON deserialization function to Community data model…](https://github.com/berkmancenter/frankly/commit/5a483680f6c0e02798dc815e581b862482c42937) so that legacy onboarding step names (which use enum values) can be migrated to use the new enum values which are expected by the frontend.
* Build files.

* N/A

## Testing this PR

Staging is not loading communities prior to this PR. Observe that communities should load as expected after this PR.

### DIFF
--- a/data_models/lib/community/community.dart
+++ b/data_models/lib/community/community.dart
@@ -50,6 +50,16 @@ enum OnboardingStep {
   createStripeAccount,
 }
 
+List<OnboardingStep> onboardingStepsFromJson(List<dynamic> stepEnumList) {
+  // Migrate legacy stored onboarding step name enum values to new names.
+  return stepEnumList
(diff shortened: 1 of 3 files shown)


katherineqian:
@karenunify just tagged you for visibility! Attempting a fix on some stored enum values that broke on staging because they became incompatible with the frontend (the frontend no longer accepted `createGuide` as an option, only `createTemplate`, but `createGuide` is what was stored on the backend).

PR #482, opened 2026-06-11
Implement mirror check, resolve several A/V issues, and new pub/sub bus

### LINKED ISSUE #11: Implement Mirror Check
In other platforms it’s pretty typical to have a “mirror check” before you enter an event or breakout. In addition to being a better user experience, it would give people a chance to make sure they have the right camera and microphone selected before entering the room.

This would greatly reduce the number of times one or more participants in a breakout have to “fix” their AV settings while the other participants see a black square.

The mirror check should appear right before participants enter a room where their camera and microphone are available.
- For **Hosted** events, this would be immediately upon entering the event.
- For **Hostless** and **Livestream** events, this would be when entering breakouts.

## Acceptance Criteria:

1. Users see a “mirror check” screen before joining that shows:
    - Live camera preview
    - Selected camera, microphone, and speaker
    - Dropdown lists for camera, microphone, and speaker selection
2. Users can:
    - Switch between available devices and see changes immediately
    - Proceed to join only after clicking an explicit "Enter breakout” button
    - (Nice-to-have) Test speakers by playing a sound
    - (Nice-to-have) Save device preferences for future sessions

### Error States

- Clear messages displayed for missing devices or permission issues
- Option to proceed without video/audio if devices are unavailable
- Instructions for resolving common device issues

### Technical Requirements

- Works across major browsers (Chrome, Firefox, Safari, Edge)
- Maintains selected preferences between sessions
- Functions on both desktop and mobile devices

johnnycrich:
## What is in this PR?

Introduces a basic "mirror check" feature to ensure users verify their camera and microphone settings before joining a meeting, as well as a lightweight publish/subscribe event bus to decouple AV device state changes from the rest of the app. The changes also enhance error handling and user feedback during device setup.

## Changes in the codebase

This pull request introduces a new "mirror check" feature to ensure users select and verify their audio/video devices before joining an event, and refactors device management to use an event bus instead of direct conference room manipulation. It also improves the UI and error handling in the media settings dialog. The most important changes are:

**Mirror Check Feature:**

* Added logic to show the `MediaSettingsWidget` as a required "mirror check" dialog when entering event; completion is tracked per event in `SharedPreferencesService` (`event_info.dart`, `shared_preferences_service.dart`).
* Added methods to `SharedPreferencesService` to persist and check if the mirror check has been completed for a given event (`shared_preferences_service.dart`).
* Introduced an `EventBus` system and `AVDeviceChangedEvent` to decouple device state changes from the conference room logic, replacing direct calls to conference room methods with event emissions (`event_bus.dart`, `media_settings_widget.dart`).
* Updated `MediaSettingsWidget` to support mirror check mode, improved UI labels, error messages, and save logic, and ensured proper device update flows and error handling (`media_settings_widget.dart`).
* Improved rendering of the video preview and error messages, and updated button labels for clarity in the device settings dialog.

**New pub/sub implementation:**

Previously, `MediaSettingsWidget` called `ConferencRoom` methods directly when the user changed devices. This created tight coupling — the settings widget had to "know about" the conference room. The changes use a classic mediator pattern — components communicate through the bus rather than holding direct references to each other, making the code easier to test and extend.

* `AVDeviceChangedEvent` — A concrete event that carries a list of changes. E.g. a single emit could signal both `disableVideo` and `updateAudioDevice` at once.
* `EventBus` — The core bus, backed by RxDart's `PublishSubject`.
* `emit(event)` — pushes an event to all current listeners
* `stream` — a `Stream<AppEvent>` that any widget/service can subscribe to
* `dispose()` — closes the subject and cleans up

**Other fixes:**

`MediaSettingsWidget`: Fixed a race condition / camera stream leak.
- A `_disposed` flag was introduced to detect when the widget is torn down while `getUserMedia()` is still in flight (e.g. user closes the dialog mid-device-switch).
- If `_disposed` is true after `getUserMedia()` returns, the newly acquired stream is immediately stopped via `stopPreviewMediaStream()` to prevent a leaked/orphaned camera stream.
- Fix in `dispose()`: it was previously calling `updatePreviewWidget()` on teardown, which would open a new camera stream that would never be stopped. That call was removed.

`AgoraParticipant`: when disabling video, explicitly stop the camera capture created by `_rtcEngine.startPreview()`.

## Testing this PR

- [ ] Create a new event and enter it. For now, use hosted.
- [ ] You should see a pop-up stating "Audiovisual Settings", with dropdowns for both camera and audio devices. When video device changes, the preview should reflect that.
- [ ] Click save and close.
- [ ] Enter the event.
- [ ] Note that your mic and camera are on and using the devices selected in mirror check.
- [ ] Access the AV settings panel as normal and verify that changing devices still works as intended.
- [ ] Refresh the event and note that you do not see the mirror check again if entering.
- [ ] Try the above steps again with both a hostless and livestream event. You see the mirror check before entering those events as with hosted.
- [ ] Verify that when you close the mirror check, the browser is not utilizing the video or microphone stream.

## PR Checklist

Copilot assisted in drafting this PR.


client/lib/features/events/features/live_meeting/features/video/presentation/views/video_flutter_meeting.dart
```diff
@@ -426,7 +426,10 @@ class _ParticipantWidgetViewState extends State<_ParticipantWidgetView> {
               child: Container(
                 padding: const EdgeInsets.all(8),
                 alignment: Alignment.center,
-                child: ParticipantGridLayout(),
+                // Only show the grid if mirror check is complete for this event
+                child: sharedPreferencesService.hasMirrorCheckCompletedForEvent(widget.liveMeetingProvider.eventProvider.eventId)
+                    ? ParticipantGridLayout()
+                    : const SizedBox.shrink(),
```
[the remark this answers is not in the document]
johnnycrich: defaultStageView is currently not exposed as an option to be changed, so the grid view is always Brady Bunch.

client/lib/core/widgets/media_settings_widget.dart
```diff
@@ -339,127 +347,123 @@ class _MediaSettingsWidgetState extends State<MediaSettingsWidget> {
...
-                                initialVideoDeviceId &&
-                            _mediaService.selectedAudioInputId ==
-                                initialAudioDeviceId)
-                        ? null
-                        : () async {
-                            final savedInitialVideoDeviceId =
-                                initialVideoDeviceId;
-                            final savedInitialAudioId = initialAudioDeviceId;
+                    text:
+                        '${context.l10n.save} ${context.l10n.and} ${context.l10n.close}',
```
katherineqian: I would localize the entire string here (i.e. "Save and close") instead of using the English "and" grammar with string concatenation (this is something that was brought up in a different PR before).

client/lib/features/events/features/live_meeting/features/video/data/providers/conference_room.dart
```diff
@@ -379,11 +386,17 @@ class ConferenceRoom with ChangeNotifier {
     // Lock this code so that different sections toggling audio will not cause race conditions.
     await _videoTogglingLock.synchronized(
       () async {
-        await _room!.localParticipant!.enableVideo(
-          setEnabled: updatedEnabledValue,
-        );
-        if (updateProvider) {
-          liveMeetingProvider.shouldStartLocalVideoOn = updatedEnabledValue;
+        try {
+          if (sharedPreferencesService.hasMirrorCheckCompletedForEvent(liveMeetingProvider.eventProvider.eventId)) {
```
katherineqian: I’m not sure this mirror check logic should go in `toggleVideoEnabled` / `toggleAudioEnabled`. My concern is that this method should represent a specific room/media operation (turning video on/off), but with the mirror-check check added here it also becomes responsible for enforcing a specific user-flow policy. That creates a less clear contract for the method, increases coupling between preflight UX and core call logic, and could increase the chance for unexpected behavior here -- as a result, would suggest moving the mirror check logic to a separate layer (similar to what's proposed in #484)

katherineqian (changes requested):
Just marking this review as "changes requested", representing our discussion on Friday about different approaches to this that might align better with your proposal in #484! Please request me for review again when ready :)

mikewillems:
@johnnycrich I made some suggestions on the code front. On the functionality front, it seems to work nicely (with 2 noted user flow issues) unless I consult the checklist. If you want it to work the way the checklist is written then we might need some modifications, but wow the visual design is well-interpreted. Some necessary changes.

### Checklist

✓ - Create a new event and enter it. For now, use hosted.

? - initially, you should not see any other participants in the meeting, whether or not they have a A/V enabled. - this is not really applicable. The mirror check doesn't have anything to do with the other participants - it shows before entering the meeting or when checking AV settings within a meeting and has no impact on other participants.

✓ - You should see a pop-up stating "Audiovisual Settings", with dropdowns for both camera and audio devices. When video device changes, the preview should reflect that. - @mwmarino gorgeous design here!
✓ - Click save and close.

X - Note that any other participants now appear, and your mic and camera are still off. - this does NOT happen. In fact, there is no option to NOT turn video on when entering a call - you can check yourself before hand but then have no option to not wreck yourself other than not joining. Camera turns on automatically, by default.

X - Verify that after turning them on your camera and microphone use your desired devices. - this doesn't always work. Haven't figured out why yet but always works in the preview, then Agora sometimes discards.

✓ - Access the AV settings panel as normal and verify that changing devices still works as intended.

✓ - Refresh the event and note that you do not see the mirror check again, and that your devices now turn on by default. - this is true, but is it what we want? I would rather have a mirror check upon each refresh / join personally.

X? - Try the above steps again with both a hostless and livestream event. You should not see the mirror check in those events before you enter a breakout room. - Why shouldn't I with hostless? I do with hostless - contradicts this bullet but I think is correct.

✓ - Verify that when you close the mirror check, the browser is not utilizing the video or microphone stream. - Indeed it stops for a brief moment before starting again once automatically joining the meeting after the mirror check.

---

Conclusion: I think this needs more UI design work. 

Imo, the mirror check should be paired with an option to not enable the camera, and especially also a way to back out and "Cancel" or "Don't Enter Yet" or "Stay Here" or something (the only option right now is "Save and Close"). Lacking those options - especially the second - is a major UI anti-pattern.

An even better flow would be that the user enters the meeting with the camera OFF by default and is immediately prompted (with mirror check) to enable it to continue in the meeting (maybe with some resistance like "Are you sure you don't want to enable your camera? Deliberation works a lot better when real people can see each other!". The flow once in the meeting (it just is included with AV settings) is great. 

Even if not having design review though, the checklist needs rewritten.

### Code Nits (not in order of priority)

1. Linter error: unused var eventId in conference_room.dart

2. I first had forgotten to reenable camera permissions for the browser I'm using, so I had no camera as far as the app was concerned. But this isn't handled gracefully. There's a null error:
```
 ⛔ Null check operator used on a null value
main.dart.js:51603 Error swallowed in tryCatch utility.
main.dart.js:51603 TypeError: Cannot read properties of null (reading 'toString')
```
Seems to come from `media_settings_widget.dart` and crashes the app / prevents loading the meeting.

3. `_onAVDeviceChanges` could be a race I think between updateAgoraVideoDevice() and disableVideo. But we'll be thoroughly addressing that later.

4. Linter: `event_bus.dart` missing newline EOF

5. Linter: `event_info.dart` had a space change on save.

6. Linter: unused import in `control_bar.dart`

client/lib/core/data/services/event_bus.dart
```diff
@@ -0,0 +1,27 @@
...
+  AVDeviceChangedEvent({this.changes = const []});
+ }
+
+// Event bus for emitting and listening to application-wide events
+class EventBus {
+  final _subject = PublishSubject<AppEvent>();
+  Stream<AppEvent> get stream => _subject.stream;
+  void emit(AppEvent event) => _subject.add(event);
+  void dispose() => _subject.close();
+}
```
mikewillems: missing EOF newline

client/lib/features/events/features/live_meeting/features/video/data/providers/conference_room.dart
```diff
@@ -481,6 +540,8 @@ class ConferenceRoom with ChangeNotifier {
   }) async {
     Debug.log('ConferenceRoom._onConnected => state: ${room.state}');
 
+    String eventId = liveMeetingProvider.eventProvider.eventId;
```
mikewillems: Linter says this is unused

client/lib/features/events/features/live_meeting/features/video/data/providers/conference_room.dart
```diff
@@ -284,6 +291,34 @@ class ConferenceRoom with ChangeNotifier {
...
     liveMeetingProvider.eventProvider.addListener(_muteOthersOnOverride);
+
+    _avDeviceChangeSubscription = appEventBus.stream
+        .whereType<AVDeviceChangedEvent>()
+        .listen(_onAVDeviceChanges);
+  }
+
+  // When AVDeviceChangedEvent is emitted on event bus, check the changes in the event and toggles the video/audio enabled state accordingly;
+  // Also updates the audio/video device if necessary.
+  void _onAVDeviceChanges(AVDeviceChangedEvent event) {
```
mikewillems, line 302: We should revisit this after building the queue / integrating Riverpod

client/lib/features/events/features/live_meeting/features/video/presentation/widgets/control_bar.dart
```diff
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:client/core/utils/toast_utils.dart';
 import 'package:client/core/widgets/media_settings_widget.dart';
+import 'package:client/features/community/presentation/widgets/community_tag_builder.dart';
```
mikewillems: Not used per linter

mikewillems (changes requested):

mwmarino:
@mikewillems @johnnycrich with default behavior (or not) with cameras and mics being on post-mirror check: ideally there should be a clear toggle that is part of the mirror check, one that indicates what will be enabled or not when you join a call. Without those toggles, I'd say it's safe to assume as a default that the camera and mic will be on. 

We need to discuss more how to proceed. As an example: 
Screenshot 2026-08-19 at 1 27 18 PM

johnnycrich:
@mikewillems Thank you for the thorough review! I made changes to address those nits and updated the testing checklist to reflect all of the most recent changes to this PR. 

- [ ] ~~Initially, you should not see any other participants in the meeting[...]~~
- [ ] ~~Note that any other participants now appear, and your mic and camera are still off.~~
- [ ] Note that your mic and camera are on and using the devices selected in mirror check.
- [ ] ~~Try the above steps again with both a hostless and livestream event. You should not see the mirror check in those events before you enter a breakout room.~~
- [ ] Try the above steps again with both a hostless and livestream event. You see the mirror check before entering those events as with hosted.

mikewillems:
Call out to a few parts of the review:

NOT FIXED / high priority: X - Verify that after turning them on your camera and microphone use your desired devices. - this doesn't always work. Haven't figured out why yet but always works in the preview, then Agora sometimes discards.

I Slacked you about this one, I think `AVDeviceChangedEvent` is missing an `appEventBus.emit()` in `media_settings_widget.dart`

---

### Code Nits (not in order of priority)

1. FIXED ~~Linter error: unused var eventId in conference_room.dart~~

2. NOT FIXED - you could say it doesn't matter because we want camera on, but aside from being anti-user, that fully crashes the app when the code always tries to unwrap a null device ID.
I first had forgotten to reenable camera permissions for the browser I'm using, so I had no camera as far as the app was concerned. But this isn't handled gracefully. There's a null error:
```
 ⛔ Null check operator used on a null value
main.dart.js:51603 Error swallowed in tryCatch utility.
main.dart.js:51603 TypeError: Cannot read properties of null (reading 'toString')
```
Seems to come from `media_settings_widget.dart` and crashes the app / prevents loading the meeting.

3. DEFER ~~`_onAVDeviceChanges` could be a race I think between updateAgoraVideoDevice() and disableVideo. But we'll be thoroughly addressing that later.~~

4. FIXED ~~Linter: `event_bus.dart` missing newline EOF~~

5. NOT FIXED but truly a nit, though I missed an unused import in the same file: Linter: `event_info.dart` had a space change on save. 

6. FIXED: ~~Linter: unused import in `control_bar.dart`~~

client/lib/features/events/features/live_meeting/features/video/presentation/widgets/control_bar.dart
```diff
@@ -41,17 +41,56 @@ class _ControlBarState extends State<ControlBar> {
...
+                shouldShowVideoPreview:
+                    !(defaultTargetPlatform == TargetPlatform.iOS ||
+                        defaultTargetPlatform == TargetPlatform.android),
+                isMirrorCheck: true,
+              );
+            },
+          );
+        }
+
+        await sharedPreferencesService.setMirrorCheckCompleteForEvent(eventId);
```
mikewillems, line 77: We need a conditional for this - right now it sets even if the user cancels, so they won't get the mirror check on their next join if they cancel. Other than that looks great!

mikewillems (approved):
Hey @johnnycrich 1 small change needed but other than that good to go! Thanks for addressing these!

PR #492, opened 2026-06-17
Add email verification to the app.

karenunify:
## What is in this PR?
One time email verification on sign-up/login if not already verified using Firebase Auth. When a user creates a new account or is logging back in, once their email is submitted, a query is made to Firebase Auth to see if that user's email has been verified. When creating the account, they see a page prompting them to check their email and verify their account with a link that is valid for 30 minutes and should not be able to get past that page until their email is verified. Same when logging back in with an existing account that has not had the email verified.

## Changes in the codebase

- Using Firebase Auth's verifyEmail(), any new and or existing user must verify their email address on login.
- Using Firebase Auth's verifyBeforeUpdateEmail(), any new or existing user can edit their email address immediately upon account creation, in the case of a typo, or an existing user who would like to update their email when verifying can do so. In the case that the email address is already in use by another account when a user edit's, because of Firebase's built-in email enumeration protection, user's will not get an error message that the address is already in use, but rather see a success message that the email was sent and it should be received in their inbox if the address is not already in use. 

* 

The app domain needs to be whitelisted in Firebase Auth, and Firebase Auth email verification template/action settings need to be modified:

1. In the Firebase console, go to Security > [Authentication](https://console.firebase.google.com/project/_/authentication/).
2. In the Sign-in method tab, enable the Email/Password sign-in method. Note that email/password sign-in must be enabled to use email link sign-in.
3. In the same section, enable the Email link (passwordless sign-in) sign-in provider.
4. Click Save.

The Email address verification email template will need to be customized in the production Firebase project.

1. Open the [Firebase Console](https://console.firebase.google.com/) and select your project.
2. In the left-hand navigation sidebar, click on Authentication.
3. Click on the Templates tab at the top of the page.

* 

## Testing this PR
Create a new account:

after signing up, a user should be redirected to a page prompting them to verify their email
they should not be able to get past this page until email has been verified, either by refreshing the page, or signing in from another tab
Log in with an existing account where the user has not verified their email:

once submitting their email/password, an existing user should be redirected to a page prompting them to verify their email
they should not be able to get past this page until email has been verified, either by refreshing the page, or signing in from another tab

## PR Checklist

- [x ] Where applicable, I have added localization (l10n) entries to my feature for user-facing text.

Claude Code extension in VS code for best practices and debugging


karenunify:
Latest push is on https://gen-hls-bkc-7627--pr492-kh-feat-email-verifi-261arhcr.web.app/

katherineqian (approved):
Looks great, thanks @karenunify! Let's get this merged (whenever you're ready) so we can test it out on Monday!

Have a great weekend!

PR #514, opened 2026-06-30
Add "last active time" and fix breakout room assignments in registration exports.

katherineqian:
## What is in this PR?

Adds a new column to the registration data exports which tells admins, roughly, when the user was last present in an event. This is also a bugfix PR that fixes the "Breakout Room Assigned" column, which was returning "Main Room" for all participants. 

## Changes in the codebase

* Pull breakout room info from the Breakout Room data list instead of from the Participant, where "currentBreakoutRoomId" gets overwritten when users leave the event (by `UpdatePresenceStatus`) 😓 
* Used local instead of UTC timestamps
* Add and export `mostRecentPresentTime` on event participants so admins can also access this data

* 

## Testing this PR

* Choose a hostless event that was used for testing breakout rooms. **Ensure that UpdatePresenceStatus fired on the participants when they left the meeting**, which clears their breakout room info and had caused the previous bug. Using an old event is a more convenient way to do this.
* Download registration data.
* Observe that prior to this PR, the breakout rooms say "Main Room" for all participants. After this PR, there should be different room numbers.
* Observe that Last Active Time is now a new column.

## PR Checklist

## Additional information

Copilot assisted in writing this PR and adding localizations.

### DIFF
--- a/client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
+++ b/client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
@@ -134,6 +134,7 @@ class _EventDataDownloadDialogState extends State<EventDataDownloadDialog> {
       await provider.generateRegistrationDataCsvFile(
         eventId: event.id,
         registrationData: members,
+        participantData: participants.toList(),
         breakoutRooms: breakoutRooms,
       );
     } finally {
--- a/client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
+++ b/client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
@@ -424,7 +424,7 @@ class _EventDataDownloadDialogState extends State<EventDataDownloadDialog> {
                   ),
                   title: Text(context.l10n.registrationDataDownload),
                   subtitle: Text(
-                    'Times are shown in your local timezone.',
+                    context.l10n.timesShownInLocalTimezone,
                     style: context.theme.textTheme.bodySmall?.copyWith(
                       color: context.theme.colorScheme.onSurfaceVariant,
                     ),
--- a/client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
+++ b/client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
@@ -423,6 +423,12 @@ class _EventDataDownloadDialogState extends State<EventDataDownloadDialog> {
                     () => registrantListSelected = value ?? false,
                   ),
                   title: Text(context.l10n.registrationDataDownload),
+                  subtitle: Text(
+                    'Times are shown in your local timezone.',
+                    style: context.theme.textTheme.bodySmall?.copyWith(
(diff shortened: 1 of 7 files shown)


client/lib/features/events/features/event_page/data/providers/event_provider.dart
```diff
@@ -409,6 +411,13 @@ class EventProvider with ChangeNotifier {
       );
       row.add(
         registrationData[i].memberEvent?.participant?.createdDate?.toUtc(),
+      final mostRecentPresentTime = participantData
+          .firstWhereOrNull((p) => p.id == registrationData[i].id)
```
mikewillems, line 418: so clean... 👏👏

client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
```diff
@@ -423,6 +423,12 @@ class _EventDataDownloadDialogState extends State<EventDataDownloadDialog> {
                     () => registrantListSelected = value ?? false,
                   ),
                   title: Text(context.l10n.registrationDataDownload),
+                  subtitle: Text(
+                    'Times are shown in your local timezone.',
```
mikewillems: You just made someone's life so much easier! 

I wonder if it should be in the event time zone instead of the client local timezone though - I fear the latter could cause confusion for people looking at the data and getting multiple conflicting versions. Organizers may well have remote employees in multiple time zones.

mikewillems: Also, it's a bit interesting that in the download the RSVP time is still in UTC while the last seen time is local - maybe we should consider having them match?

mikewillems (approved):
See comments on time zone, but otherwise looks great! I did find an unrelated bug in the merged code that is unrelated to this PR which I'll log - that's why it was taking me a bit to review. Awesome job @katherineqian on the rapid fix!

mikewillems (dismissed):
My code, merged into this, needs a quick change - I'll go ahead and make that and re-push

mikewillems (approved):

client/lib/features/admin/presentation/widgets/event_data_download_dialog.dart
katherineqian: Thanks Mike! Since the time of the event displayed on the app is shown in local time, I figured it'd make sense to have it match for now. We could always change it if we get feedback from organizers that it's confusing?

Also, I double checked and the RSVP time should have been updated to UTC in the same [commit](github.com/berkmancenter/frankly/pull/514/commits/6ef38b847a07bc30120c7d03c4a443b8348d7dbf) that I changed the last present time parsing. I also tested a download and can't repro a discrepancy. Can you confirm whether the two columns are different?

katherineqian:
@mikewillems Can you actually separate out the fix for the OnEventParticipant write, since it's not directly related to this PR?

Due to being a hotfix, I realized this fix actually needs to branch off main rather than staging, so I will be creating a separate branch with cherry picked commits from this one. Thanks for dealing with the coordination here!

PR #535, opened 2026-07-14
Fix renderer inconsistency between local and preview vs prod and staging deployments

mikewillems:
The deploy-preview and test_client workflows were passing 'web-renderer html' as positional arguments (missing -- prefix), causing Flutter to silently ignore them and build with the default 'auto' renderer instead of 'html'. This made preview/test builds render with CanvasKit on desktop while prod/staging explicitly use the HTML renderer.

This means:

- We can get rendering differences (shadows & clipping e.g.) between preview and prod
- Preview builds are not accurately representing what will ship to production (and e.g. include extra WASM payload)

With this fix, all environments use the HTML renderer, making preview match prod deployments.

## What is in this PR?

- Fixes the `--web-renderer html` flag syntax in two CI workflow files so the HTML renderer is explicitly selected (matching prod/staging behavior).

## Changes in the codebase

- `.github/workflows/deploy-preview.yml`: Changed `flutter build web --release web-renderer html` -> `flutter build web --release --web-renderer html`
- `.github/workflows/test_client.yaml`: Changed `flutter build web --release web-renderer html` -> `flutter build web --release --web-renderer html`

- None

## Testing this PR

- Verify preview in Chrome DevTools -> Network that CanvasKit isn't downloading (no `canvaskit.wasm` request).
- Alternatively, look at the GitHub Actions logs and see that `--web-renderer html` being parsed correctly and the right renderer is being used (it's logged HTML instead of canvaskit or WASM).

## PR Checklist

- [NA] Where applicable, I have added localization (l10n) entries to my feature for user-facing text.
- [NA] For new Cloud Functions, I have added the function to `function-mapping.json`.

## Additional information

- The bug was harmless in that builds still succeeded - Flutter's arg parser treats the unrecognized positional arguments as noise and continues with defaults.

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -16,15 +16,7 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v3
 
-      - uses: dorny/paths-filter@v3
-        id: changes
-        with:
-          filters: |
-            client:
-              - 'client/**'
-
       - name: Set Up Flutter
-        if: steps.changes.outputs.client == 'true'
         uses: subosito/flutter-action@v2
         with:
           flutter-version: "3.22.2" # could read version from pubspec file, but would need to be exact
@@ -36,21 +28,17 @@ jobs:
           pub-cache-path: "/home/runner/.pub-cache"
 
       - name: Install Dependencies
-        if: steps.changes.outputs.client == 'true'
         run: flutter pub get
 
       - name: Update Product Version
-        if: steps.changes.outputs.client == 'true'
         run: |
           sed -i 's/__VERSION__/${{ github.sha }}/g' ${{ github.workspace }}/client/web/index.html
 
       - name: Update Google Auth Client ID
-        if: steps.changes.outputs.client == 'true'
         run: |
           sed -i 's/__GOOGLE_ID__/${{ secrets.GOOGLE_SIGN_IN_ID }}/g' ${{ github.workspace }}/client/web/index.html
 
       - name: Create .env file
-        if: steps.changes.outputs.client == 'true'
         uses: SpicyPizza/create-envfile@v2.0
         with:
           envkey_FIREBASE_API_KEY: ${{ secrets.FIREBASE_API_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-preview.yml
```diff
@@ -16,15 +16,7 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v3
 
-      - uses: dorny/paths-filter@v3
```
johnnycrich, line 19: @mikewillems I think we still want to keep this step so we don't have needless build generation, which uses billable resources. There may also be cases where a changeset is small enough that kicking off another action within the workflow is superfluous.

I like your idea about forcing a build when needed, though. One thing we might consider is adding a custom flag to the top of this workflow. Ex:

```
workflow_dispatch:
    inputs:
      force_preview_build:
        description: 'Internal use only. Do not check this box.'
        type: boolean
        required: false
        default: false

```

And then this `changes` step would be conditional on that input being false or undefined:
`if: ${{ github.event.inputs.force_preview_build != 'true' }}`

And then you can force build by using the flag in the CLI:
`gh workflow run [.....] -f force_preview_build`

johnnycrich (changes requested):
Just see my one comment, then will be good to go!

.github/workflows/deploy-preview.yml
mikewillems: Good catch! I agree that `force_preview_build` is the right approach. I'll push a quick update.

mikewillems: Okay, pushed a change - gh actions is a little particular about actions only being able to run actions (e.g. dorny/paths-filter) as async steps inside jobs - it can only gate on the results of external jobs that have finished, so I split the check off into another job instead of repeating the check before every build step. This is easier to modify later and reflects the logic of what we're doing (we're deciding at the whole job level).

Now we get a checkbox that we can use when running the action from the gh Actions interface too and have that option in addition to CLI.

johnnycrich (approved):
Good idea to run as separate jobs in the CI! Good to go!

PR #550, opened 2026-07-22
Ghost user cleanup v2!

mikewillems:
# Stale participant cleanup and presence query hardening

Supercedes #303. Builds on #297 (which established heartbeat and `mostRecentPresentTime` writes).

## What is in this PR?

Nothing in our back end detected when heartbeat updates stopped. Users who disconnected without explicitly leaving still showed up as present participants. Room-based queries (`breakoutRoomParticipantsStream`, `VoteToKick`) did not filter by `isPresent`. The `UpdateLiveStreamParticipantCount` pre-check window excluded ongoing events which froze livestream counts.

This PR adds a scheduled / interval-based cleanup function, improves query filters, fixes the livestream count, fixes participant count displays, and reduces heartbeat frequency by 4x (12x -> 3x / minute) for cost savings.

### 1 -- 2a30629e: Add composite indexes for stale participant cleanup and room queries

- `isPresent + mostRecentPresentTime` (collection group) for the cleanup query
- `currentBreakoutRoomId + isPresent` (collection) for filtered room queries
- Deployed first so indexes build before dependent code ships

### 2 -- 31c7f582: Filter `breakoutRoomParticipantsStream` by `isPresent`

- Disconnected participants no longer appear in admin panel room counts or participant lists

### 3 -- 317dc3f3: Filter `VoteToKick` participant query by `isPresent`

- Ghost participants no longer inflate the voting population denominator

### 4 -- 386946c0: Add `CleanupStaleParticipants` scheduled function

- Runs every minute via Cloud Scheduler, 6 internal passes at ~10s intervals
- Queries `WHERE isPresent == true AND mostRecentPresentTime < (now - 45s)`
- Transaction-based with freshness guard (re-reads before writing)
- Sets `isPresent: false`, preserves `currentBreakoutRoomId`

### 5 -- 8c010407: Tests for `CleanupStaleParticipants`

- Stale marked offline, fresh preserved, already-offline untouched, boundary conditions

### 6 -- 4a331ebf: Reduce heartbeat to 20s + add `onOnline` reconnect listener

- 75% fewer Firestore writes
- Immediate heartbeat on network reconnect prevents false-positive stale detection
- 45s stale threshold derived from: 2 x 20s + 5s buffer

### 7 -- 82e17973: Fix `UpdateLiveStreamParticipantCount` pre-check window

- Previous window `[now, now + 1 day]` excluded running events
- Changed to `[now - maxDuration * 1.5, now + 1 hour]` with named constants

### 8 -- 06794712: Fix count displays on admin panel, event cards, participants dialog

- `EventProvider`: `presentParticipantCount` getter filters by `isPresent`
- Admin panel, event cards, participants dialog use correct counts

### 9 -- 40b43b12: Tests for VoteToKick `isPresent` filter + participant count pre-check window

- Kick threshold scales with present participants only
- Pre-check window includes currently-running events

## Revision Commits

These commits address automated suggestions from Co-Pilot which are all solid ideas.

### 10 -- feb724a1: MembershipStatus enum value corrected to '.moderator' not '.mod'

- Fixes compilation error in vote_to_kick_test.dart

### 11 -- 71390742: Limit cleanup queries to 50 and double timeout to 120s

- Prevents excessive parallel transactions under large stale sets
- Extra timeout buffer for safety

### 12 -- 8860df8a: Fix kArrowLeftThinSvg isSvg flag

- SVG asset needs isSvg=true for correct widget selection

### 13 -- 7feee4ed: Make isLive consistent in participants dialog list builder

- Use hasPresentParticipants (estimate-aware) to match title logic

### 14 -- 8f115ada: Fix pre-check comment typo in count test

- Comment incorrectly said "included" instead of "excluded"

## Presence timing constants

I'll be adding this also to our architecture docs.

| Component                 | Value | Rationale                           |
| ------------------------- | ----- | ----------------------------------- |
| Client heartbeat interval | 20s   | 4x reduction from 5s                |
| Stale threshold (cleanup) | 45s   | 2 x 20s heartbeat + 5s buffer       |
| Cleanup check interval    | ~10s  | 6 passes per Cloud Scheduler minute |
| Livestream count window   | ~40s  | 15s interval + 25s heartbeat buffer |
| Pre-check lower bound     | -360m | maxEventDuration(240) x 1.5         |
| Pre-check upper bound     | +1hr  | Pre-event window                    |

## Interaction with existing cleanup (UpdatePresenceStatus)

`UpdatePresenceStatus` is RTDB-based, and we still keep it with this PR after incorporating the CF-based detection. That scheduled cleanup is a fallback if RTDB presence detection gets delayed 30-120s or skipped entirely, e.g. if the process is killed on device or the network disconnects.

If RTDB runs first, cleanup skips (already offline). If cleanup runs first, RTDB just sets offline again.

The major difference is that `UpdatePresenceStatus` clears `currentBreakoutRoomId` to null. Scheduled cleanup preserves it. All room queries now filter by `isPresent` though, so both are safe.

## Disconnect detection

This now happens in 3 layers which should cover all situations. Making some timing diagrams for the architecture docs which should clarify this.

| Layer              | Mechanism                      | Latency | Covers                                           |
| ------------------ | ------------------------------ | ------- | ------------------------------------------------ |
| 1. Explicit write  | `onBeforeUnload` / `dispose()` | none! | Tab close, navigation, explicit leave            |
| 2. RTDB disconnect | `UpdatePresenceStatus`         | 30-120s | Network loss, browser crash (SDK detects)        |
| 3. Scheduled       | `CleanupStaleParticipants`     | 45-55s  | Everything else (kill, network loss bypass RTDB) |

None.

## Testing this PR

- Unit tests cover cleanup function, VoteToKick filter, and count pre-check window
- Manual: join event, kill browser, confirm offline within 55s
- Manual: breakout room counts exclude disconnected users
- Edge case: reconnect during cleanup window (make sure onOnline restores before cleanup fires)
- Edge case: mobile background < 45s (ensure not falsely cleaned up)
- Edge case: 3+ hour event - ensure still gets count updates

## Additional Information

Incorporated revisions to the initial PR as suggested by Co-Pilot.


client/lib/features/events/data/services/firestore_event_service.dart
```diff
@@ -16,6 +16,17 @@ import 'package:rxdart/rxdart.dart';
...
 
+  // How far before an event's scheduled time it starts appearing in upcoming
+  // event queries — keeps recently-started events visible in listings.
+  static const Duration _upcomingEventsLookback = Duration(minutes: 15);
+  // Wider lookback used when querying across all communities, to surface
+  // events that started up to an hour ago.
+  static const Duration _allCommunitiesEventsLookback = Duration(hours: 1);
+  // Throttle on the event-participants Firestore stream to avoid processing
+  // a snapshot on every keystroke / write during busy periods.
+  static const Duration _participantStreamSampleTime =
```
johnnycrich, line 27: Thank you for these semantic naming updates! Love to see it.

client/lib/features/events/features/live_meeting/data/providers/live_meeting_provider.dart
```diff
@@ -204,11 +224,41 @@ class LiveMeetingProvider with ChangeNotifier {
 
   String? get breakoutRoomOverride => _breakoutRoomOverride;
 
+  /// Whether the user is currently transitioning into a breakout room. True from the moment the
+  /// user is assigned to a room until they have successfully joined the room stream.
+  bool get isInBreakoutTransition => _inTransitionToBreakoutRoomId != null;
```
johnnycrich, line 231: Thanks for things like this. Really helps to handle any edge cases/races. 😄

client/lib/features/events/features/live_meeting/features/live_stream/presentation/widgets/url_video_widget.dart
```diff
@@ -53,10 +53,15 @@ class UrlVideoWidget extends StatefulHookWidget {
   }) : super(key: key ?? Key(playbackUrl));
 
   @override
-  _UrlVideoWidgetState createState() => _UrlVideoWidgetState();
+  State<UrlVideoWidget> createState() => _UrlVideoWidgetState();
```
johnnycrich, line 56: Love all the continued cleanup here!

client/lib/features/events/features/live_meeting/features/live_stream/presentation/widgets/url_video_widget.dart
```diff
@@ -125,12 +130,13 @@ class _UrlVideoWidgetState extends State<UrlVideoWidget> {
     useEffect(
       () {
         final subscription = html.window.onMessage.listen((event) {
-          
+          if (event.origin != html.window.location!.origin) return;
```
johnnycrich: Nice catch here. I think a comment might be nice explaining the purpose of the guard.

firebase/functions/lib/events/live_meetings/cleanup_stale_participants.dart
```diff
@@ -0,0 +1,154 @@
...
+      staleParticipants.documents.map(
+        (doc) => firestore.runTransaction((transaction) async {
+          final freshSnapshot = await transaction.get(doc.reference);
+          final participant = Participant.fromJson(
+            firestoreUtils.fromFirestoreJson(freshSnapshot.data.toMap()),
+          );
+
+          // Guard: if the participant has sent a heartbeat since our query
+          // ran, they are alive and we should not mark them offline.
+          if (participant.isPresent != true) {
```
johnnycrich, line 105: Nice.

firebase/functions/lib/events/live_meetings/update_live_stream_participant_count.dart
```diff
@@ -15,10 +15,48 @@ class UpdateLiveStreamParticipantCount implements CloudFunction {
...
+  // Firestore query window. The precise per-event filter in
+  // _updateEventParticipantCount uses the actual durationInMinutes from each
+  // event document; this constant just needs to be >= the longest plausible
+  // event so the pre-check doesn't exclude currently-running events.
+  static const int _maxEventDurationMinutes = 240; // 4 hours
+
+  // The per-event active window: [scheduledTime - _preEventWindowHours,
+  //                                scheduledTime + durationInMinutes * _postEventDurationMultiplier]
+  static const int _preEventWindowHours = 1;
+  static const double _postEventDurationMultiplier = 1.5;
```
johnnycrich, line 28: How did you arrive at this value?

johnnycrich (approved):
@mikewillems I am improving in earnest, but just see one little tiny _tiny_ nit and question. 

Also did quite a bit of manual testing with the following good results:
- Joined event, left via tab close, confirm user marked offline within ~1 min
-- and breakout room count updated to reflect
- Same as above, rejoined within 20s, heartbeat marked/sent
- Saw error messaging about network error/closed breakout room when trying to enter one that ended

client/lib/features/events/data/services/firestore_event_service.dart
mikewillems: 🙏

client/lib/features/events/features/live_meeting/data/providers/live_meeting_provider.dart
mikewillems: Thanks! Yea this was SO confusing to me when I was first starting this. We had no clear destination for breakout transitions and it was pretty chaotic. This should translate nicely to a centralized action/event bus: basically the pattern is start atomizing transitions by including all the necessary info for it in the call and ensure we don't use context vars for ANYTHING but visual state.

firebase/functions/lib/events/live_meetings/update_live_stream_participant_count.dart
mikewillems: It's two distinct values, but both were somewhat arbitrary judgment call on my end. They seemed reasonable but could be named more clearly I guess - this is just the rule that defines when the backend participant counting function will run for an event.

The first value is the start time for the window of counting (and showing) participant count for the event, which I figured an hour before is plenty of time. For sure I think that before that you wouldn't want to see who was or wasn't online, but this could be moved to a lower value too. Before that the count doesn't get written (null) and the count widget is hidden.

The second value is how long after the start of an event a user can log on and still update the counts. I think some time after is good because we're not hard-ending events yet and Sam and I had talked a lot about a grace period after. I thought 50% after scheduled event end would be a good upper bound.

Open to feedback on this.

client/lib/features/events/features/live_meeting/features/live_stream/presentation/widgets/url_video_widget.dart
mikewillems: Thanks! Turns out the nullsafe operator is not necessary (I put it in without checking), so I'll remove that, but the overall guard is a security measure to prevent postMessage from a different origin (e.g. iframe). I'll comment that.

PR #584, opened 2026-08-11
Architecture Documentation - DOCS ONLY

mikewillems:
## What is in this PR?

This  PR adds only a bunch of architecture documentation for navigating the code base and doing quick lookups of where functionality lives (as well as diffing with proposed documentation). This took me months to compile, so hopefully it saves some time for future devs.

It is intended to be a working reference point rather than an authoritative spec - I propose we keep it updated on a best-effort basis but don't over-index on it. Maybe in the future we can set some flags to detect when relevant parts of the code base are changed (similar to how we do function mapping to figure out what rebuilds).


docs/pages/architecture/meeting-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+sequenceDiagram
+    participant Host
+    participant Server
+    participant Firestore
+    participant Participant
+    participant AgoraRoom
+
+    Host->>Server: InitiateBreakouts(numPerRoom, method)
+    Server->>Firestore: Set breakoutRoomStatus=pending
+    Firestore-->>Participant: LiveMeeting snapshot (pending)
```
johnnycrich: @mikewillems Syntax issue here, prevent diagram render.

johnnycrich (changes requested):

docs/pages/architecture/meeting-lifecycle.md
mikewillems: Thanks for pointing this out @johnnycrich! My local note app is a bit more permissive and let me get away with some sloppiness and rendered for me - new push should work! I think the issue was that participant is a reserve word and I was doubling up and using it as content.

johnnycrich (approved):
@mikewillems Looking great, merge away!

PR #469, opened 2026-06-09
Add configurable Firebase function regions

dtrn2048:
Summary: make Firebase callable and function regions configurable, preserve current defaults when unset, and add region-aware Hosting and CI plumbing. Validation: git diff check, node syntax checks, targeted Dart analyze, targeted Flutter analyze. Notes: opened from the Dembrane fork only to isolate the EU extraction; broader repo test and toolchain issues outside this diff still exist, so this PR is a draft.

Issue #484, opened 2026-06-12
Decouple device handling from Agora and create video interface abstraction layer


johnnycrich:
Here is a breakdown of proposed changes, written with the help of Claude:

# Decouple Device Handling & Video SDK Abstraction

## Problem

The platform relies purely on Agora for accessing and setting defaults for camera and microphone. This causes issues primarily around device configuration from a cold-start within an event. Other architectural issues result in several UX concerns, such as participants seeing or hearing other people or themselves before they have enabled A/V or while they are connecting. This creates a confusing or even *intimidating* and privacy-violating environment.

User preferences related to media devices should be handled by the platform itself, not within a third-party SDK. 

We also want consumers to be able to drop in alternative video backend SDKs (Jitsi, Zoom, etc.) someday.

---

## Current State (Problems)

The coupling exists at three levels:

1. **Device enumeration**: `MediaDeviceService` already uses `navigator.mediaDevices` (web APIs) — this part is OK — but the preview and device-apply path runs back through `ConferenceRoom → AgoraParticipant → RtcEngine`, entangling device UI with Agora's lifecycle.

2. **Mirror check is post-channel-join**: `ConferenceRoom` shows the mirror dialog *after* `onJoinChannelSuccess`. Users are already technically in the meeting before they can check their devices. **In our current basic mirror check**, we are just hiding various aspects of the UI artificially before the check is completed.

3. **No SDK abstraction**: `AgoraRoom` directly calls `createAgoraRtcEngine()`, `RtcEngineEventHandler`, `AgoraVideoView`, `VideoViewController` — nothing is behind an interface.

---

## Target Architecture (3 Layers)

```
┌─────────────────────────────────────────────────────────┐
│  Layer 1: Platform Device Layer (no SDK dependency)     │
│  MediaDeviceService + MediaDevicePreferences         │
│  Uses: camera_web / dart:html (browser MediaDevices API)│
└─────────────────────────────────────────────────────────┘
           │ device list, preview stream
           ▼
┌─────────────────────────────────────────────────────────┐
│  Layer 2: Video SDK Abstraction                         │
│  VideoSdkAdapter (abstract)                             │
│  AgoraVideoSdkAdapter (concrete)                        │
│  XXVideoSdkAdapter, YYVideoSdkAdapter (future)          │
└─────────────────────────────────────────────────────────┘
           │ events, participants, video views
           ▼
┌─────────────────────────────────────────────────────────┐
[...]

PR #612, opened 2026-09-05
Add configurable EU deployment and optional Dembrane integration

dtrn2048:
## What is in this PR?

Allow self-hosted Frankly installations to select a Firebase function region and optionally connect recorded events to a Dembrane Echo project. The default remains the existing region with Dembrane disabled. This combines the EU deployment capability proposed in #469 with the optional integration on the current staging foundation.

## Changes in the codebase

- Configure callable functions, Cloud Tasks, Hosting rewrites, and client endpoints consistently for the selected region.
- Gate Dembrane linking, branding, recording consent, automatic recording, and recording forwarding behind client and server flags that default to false.
- Forward completed recording artifacts to a configured authenticated bridge, retain successful artifact markers, and report delivery failures without blocking transcript processing.
- Preserve current upstream transcription and matching behavior. Route root HTML through ServeIndex and fix template lookup for both function entrypoints and skip notification recipients without email addresses.
- Fail the client build when `AGORA_APP_ID` is missing, so a web build cannot ship without it.
- Ignore per-environment `.env.<env>` and `.runtimeconfig.<env>.json` files, and the Firebase extension params file, so local credentials stay out of git.
- Add `CleanupStaleParticipants` to `function-mapping.json`.
- Document configuration, deployment, and rollback.

Opt-in deployments set `DEMBRANE_ENABLED=true` for the client and `dembrane.enabled=true`, `dembrane.bridge_url`, and `dembrane.bridge_token` for functions. Bridge credentials remain server-side. The separately hosted bridge is not deployed by this repository.

## Testing this PR

- Full Chrome client suite: 561 passed, 20 skipped.
- Flag-enabled client checks: 45 passed, 1 skipped; disabled: 42 passed, 4 skipped.
- Backend Firestore emulator suite: 143 passed; subsequent notification regression checks: 9 passed.
- Firestore rules: 52 passed. JavaScript region, bridge, and config-renderer checks passed.
- Client release and backend builds passed. All 78 function exports load with production configuration.

Production rollout: all 78 app functions are active in europe-west1. Live private-community checks passed for anonymous authentication, event creation, joining, Agora token issuance, and event editing. The deployed browser frontend loads with Dembrane branding and no observed startup console errors. The production root returns a matching CSP nonce and no unresolved configuration placeholders. Production artifact-to-Echo forwarding passed using a synthetic three-second MP4 tone: the stopped-session trigger located the artifact, the existing bridge converted and uploaded the audio and finished the Echo conversation, and Frankly recorded delivery success with no post-processing error. A live meeting capture test then passed end to end: a real browser join recorded 6.3 MiB of audio to Cloud Storage as MP4 plus segments, Agora real-time transcription produced VTT with correct speech and speaker attribution, and both the recording and the transcripts were registered on the recording session. Existing production membership rules are retained by explicit deployment-owner decision.

## PR Checklist

- [x] For new Cloud Functions, I have added the function to `function-mapping.json`.

## Additional information

Artifact success markers prevent routine repeat delivery; they are not a transactional cross-service idempotency guarantee. Free-text smart matching still depends on the optional hosted matching API; local fallback supports binary survey responses.

**AI tools used**: OpenAI Codex assisted with implementation, conflict resolution, tests, deployment investigation, and this description. Human review is requested.


saihla:
@dtrn2048 Thank you for this! A member of our team will take a look at this in the coming days and respond with next steps. Excited about this integration!

johnnycrich:
Thanks for the thorough work here! The region-configurability piece is genuinely useful for any self-hoster.

I would like to bubble a concern before this goes further, though: the Dembrane integration itself is architected in a way that couples core Frankly code to one specific toolset, and that's a direction we're actively trying to move away from. A few concrete examples:
- `dembraneProjectId` and `hasDembraneProjectLink` are now first-class fields on the shared `Event` model (`data_models/lib/events/event.dart`)
- `DembraneProjectLinkSection` is a partner-named widget living in the general `events/presentation/widgets` folder
- `dembrane-bridge.js` puts partner-specific forwarding logic directly in our Cloud Functions

This would mean every fork/self-host of Frankly now carries Dembrane-specific data fields, UI, assets, and function code, whether or not it uses Dembrane. We're trying to get away from this pattern — see #484 as an example.

What we'd propose instead: keep Frankly's side generic. Rather than a Dembrane-shaped field/widget/function, add a general-purpose capability — e.g. an event/community-level "external delivery" config. This might fire when a recording/transcript is ready. Your existing frankly-dembrane-bridge is a natural home for the Dembrane-specific half of this — translating that generic payload into Echo API calls, etc. That keeps your logic in your control, and it means the same mechanism can support other partners/self-hosters later without trunk changes.

Two suggestions on how to move forward:
1. We'd be happy to have the region-configuration piece as its own PR — that piece seems ready for a merge roughly as is.
2. For the integration piece, we'd like to collaborate on the shape of a generic webhook/payload contract, and have the Dembrane-specific translation live in your bridge. We can align on data shape in an issue/discussion if that's useful. 

I want this to work well for you, but just want to make sure we land somwhere that works long-term for both of us and for anyone else self-hosting.

Please let me know your thoughts!

PR #619, opened 2026-09-10
Fix participant count display, campground fix

johnnycrich:
## What is in this PR?

User interface fixes / consistency improvements in the `ParticipantsList` widget, and makes a minor update to a workflow configuration file (campground). Ensures that participant count labels are properly localized and that edge cases (such as a single participant with no available name) are handled gracefully.

## Changes in the codebase

* Added a specific case for displaying `'1 person'` (localized) when there is a single participant and the participant list is empty.
* Replaced hardcoded strings like `'1 person'` and `'<n> People'` with localized versions using `context.l10n.person` and `context.l10n.people` throughout the widget. 
* Updated the import statement to use `package:provider/provider.dart` instead of the internal path, following best practice.
* Removed the mapping for `AdvanceMeetingGuideAfterDelayServer` from `.github/workflows/function-mapping.json` as its not a real function target.

## Testing this PR

- Create an event and make it hostless.
- Change the event to "hosted" in the edit settings panel.
- Note that there is no longer an errant box, and participant max counts are accurate.

## PR Checklist

Copilot helped drafting this PR ticket.


.github/workflows/function-mapping.json
```diff
@@ -203,9 +203,6 @@
     "firebase/functions/lib/events/live_meetings/breakouts/advance_meeting_guide_after_delay.dart": [
       "AdvanceMeetingGuideAfterDelay"
     ],
-    "firebase/functions/lib/events/live_meetings/breakouts/advance_meeting_guide_after_delay_server.dart": [
```
mikewillems, line 206: Why do you think this is unused?

mikewillems (commented):
Unfortunately I accidentally directly pushed a fix for this stuff, but also I think this deletes a needed function registration. Shall we close?

johnnycrich:
@mikewillems I removed `AdvanceMeetingGuideAfterDelayServer` from mapping because I attempted to deploy that function manually, and it failed because it is not an exported function; it is just an internal helper. So future CI-triggered deploys for that target would fail. Let me know if you think there is a different approach. Yes, I'm going to close.

PR #622, opened 2026-09-11
Mirror the "mirror-check" preview

mikewillems:
I had to know... turns out it was literally a CSS one-liner :P

That's all this PR does. No back end or significant changes or side effects... test that it works and we're done 🥳

The only question is whether we want the user's self-view to be likewise mirrored in the actual meeting... I think yes, but that's not in this quick fix. That's just another one-liner: `mirrorMode: VideoMirrorModeType.videoMirrorModeEnabled`. @johnnycrich and @saihla pls comment in the review if you think we should do that.


johnnycrich (approved):
Nice!

mikewillems:
@johnnycrich as a heads up, I went ahead and mirrored for the in-meeting settings view also, since that was also a 1-liner (applied to 2 files since there are separate settings dialogs for the 2 layouts).

PR #623, opened 2026-09-12
only show participants section for hosted event in settings drawer

mikewillems:
Just a tiny quick fix to the random user image where the max participant count would be in the event settings drawer.


mikewillems:
@johnnycrich  this is a tiny fix to the already-approved PR that I forgot to stick on there. Just a quick sanity check and we can finish out the feedback from today!

johnnycrich (approved):
@mikewillems You know how much I love a tiny PR! Good to go!

PR #632, opened 2026-09-18
Mw/fix/timer display

mikewillems:
## What is in this PR?

This is a fix for the agenda item countdown timer not showing up until later than it should much of the time and not showing up for certain types of agenda item. This was 🐢 on 🐢 on 🐢 . I wrote thorough commit messages in lieu of a thorough PR message, but let's see how well Claude can imitate me 👁️‍🗨️🙏👁️‍🗨️    

## Changes in the codebase

This PR fixes three defects in the meeting-guide advance countdown ("moving onto the next agenda item" ring) and locks in the backend advance timing with a test. The advance flow is backend-driven: a client ready-vote triggers `ParticipantAgendaItemDetailsOnWrite`, which schedules `AdvanceMeetingGuideAfterDelay` after `meetingGuideAdvanceDelay` (8s); the client renders the countdown ring while an advance is pending. The bugs were: (1) the mobile ring never appeared because it gated on a value that is transiently null, (2) in the emulator the scheduled advance was silently dropped due to a project-id URL mismatch, and (3) the ring briefly flashed on the final agenda item, which has no next item to count down to.

The work is split into four commits:

### 1. `Fix mobile advance-countdown null-id suppression; add gate tests` (`09f14b76`)
The root cause of the "countdown never appears on mobile" case: the mobile bottom nav gated the ring on `getCurrentAgendaItem()?.id`, which is `null` whenever the displayed agenda-item object hasn't resolved yet, so a genuinely active server advance for the current item was not recognized and the ring was suppressed.

- **`client/lib/.../presentation/views/live_meeting_mobile_page.dart`** — Resolve the current id once via `_presenter.getCurrentAgendaItemId()` (which returns the pending item id safely) and pass it to both `MeetingGuideCardStore.detailsForAgendaItem(...)` and `isPendingAdvanceOptimistic(...)`, replacing the null-prone `currentItem?.id`.
- **`client/test/lib/.../meeting_guide_card_presenter_test.dart`** — New presenter gate tests covering `isPendingAdvanceOptimistic` (server-pending match, optimistic-threshold, below-threshold, and the null-id suppression case) and `getCurrentAgendaItemId()` resolving the pending item when `getCurrentAgendaItem()?.id` is null.
- **mocked_classes.mocks.dart** — Regenerated/added mock support (`MockMeetingGuideCardStore` and related fakes) required by the new tests.

### 2. `Add backend advance-timing test` (`51ae007c`)
Locks in the server-side countdown timing so a regression in the delay is caught.

- **check_advance_meeting_guide_test.dart** — New test asserting the scheduled advance fires ~`meetingGuideAdvanceDelay` (8s) out and not sooner.

### 3. `Fix emulator scheduled-call URL to target the running emulator project` (`1ceefed4`)
In the emulator, `ScheduledFunctions` posts the direct callback to the configured `app.functions_url_prefix`, whose project id can differ from the project the emulator was started as (`GCLOUD_PROJECT`). The mismatch 404s and the scheduled call is silently dropped, so pending advances never fire.

- **emulator_utils.dart** — Add `String? get emulatorProjectId`, reading `GCLOUD_PROJECT` from the process env and returning null when unset (deployed env / unit tests).
- **scheduled_functions.dart** — In the emulator branch, compute `_emulatorUrlPrefix(urlPrefix)` and post to it instead of the raw prefix. New top-level `_emulatorUrlPrefix` helper rewrites the project-id path segment to `emulatorProjectId`, falling back to the original prefix when the project id is unknown or the prefix is malformed. The URL is rebuilt as an explicit string rather than via `Uri.replace(pathSegments:)`, which drops the scheme when compiled to JS (producing a `Protocol ":" not supported` error in the node http client).

### 4. `Suppress countdown timer on final agenda item` (`46201d67`)
The backend finishes the meeting immediately on the last agenda item (`finishMeeting`) with no pending-advance window, so the countdown ring has nothing to count down to and should not flash up.

- **`client/lib/.../meeting_guide/presentation/meeting_guide_card_presenter.dart`** — Add `isLastAgendaItem(String?)` (true when the id equals the last of `_agendaProvider.resolvedAgendaItems`) and return `false` early from `isPendingAdvanceOptimistic(...)` for the final item.
- **`client/lib/.../presentation/live_meeting_mobile_presenter.dart`** — Same `isLastAgendaItem` helper and early-return guard, so the final-item behavior matches on mobile.

**No user-facing strings and no `.arb`/l10n changes** are introduced. The final-item change only conditionally suppresses an already-localized widget.

None. 

## Testing this PR

- Create a meeting with more than the one default agenda item.
- Go through the breakouts agenda once, all within the first few seconds; and verify advance works as expected (with timer) even with warning, and that canceling stops the timer and advance.
- Restart breakouts and try again, this time waiting 30s per item. Verify it works as expected.
- In both cases, verify that the last item doesn't show a timer (even briefly) and just concludes immediately.

## PR Checklist

- [x] I have built locally and committed any updated freezed files.
- [NA] For new Cloud Functions, I have added the function to `function-mapping.json`.
- [x] I have ensured that all touched files have been formatted and are unmarked by the linter.
- [x] I have updated any relevant documentation for changed functionality.

## Additional information

There are problems with entering breakout rooms. I'm writing a ticket for that. So don't judge the effectiveness of the timer fix on the entry.

Claude Sonnet 4.8 to write the above code summary.


docs/pages/dev.md
```diff
@@ -172,6 +172,8 @@ Documentation for the emulator suite is available here:
 
 The emulators run against the Firebase project ID specified by `FIREBASE_PROJECT_ID` in `client/.env`. This must match your actual Firebase project. If you see CORS errors when trying to load pages in your local version, double check that this has been set correctly.
 
+> **Scheduled functions in the emulator:** Cloud Tasks can't dispatch callbacks to a local `127.0.0.1` emulator, so scheduled work (e.g. the delayed agenda advance) is invoked by a direct HTTP callback instead. The callback URL is rewritten to the project the emulator is actually running as (`GCLOUD_PROJECT`), so a configured `app.functions_url_prefix` that points at a different project id won't silently 404 and drop the call.
```
johnnycrich, line 175: 🙌🏻

firebase/functions/lib/events/live_meetings/breakouts/check_advance_meeting_guide.dart
```diff
@@ -296,7 +298,16 @@ class CheckAdvanceMeetingGuide {
     // actually triggering the advance after the delay.
     return AdvanceCheckResult(
       isPendingOrAdvancing: true,
-      isLastAgendaItem: currentAgendaItemId == event.agendaItems.lastOrNull?.id,
+      // Use the resolved agenda (same as advanceMeetingGuide's navigation) so we
+      // don't use timers for last-items even when they are `{diffusionStatement}`
```
johnnycrich: I don't think "even when they are `{diffusionStatement}`" Is really relevant here. As in, that would never change the condition, so not really worth noting, even though you are using its resolver as a right-hand comparator. "Don't use the timer for last items, need to use ID returned by resolved items.", or something akin.

firebase/functions/lib/utils/infra/scheduled_functions.dart
```diff
@@ -71,11 +71,18 @@ class ScheduledFunctions {
       print(
         'Emulator detected: calling $functionName directly instead of via Cloud Tasks',
       );
+      // The emulator serves functions under the project id it was started with
+      // (`--project`, exposed as GCLOUD_PROJECT). The configured
```
johnnycrich, line 75: Nice catch here!

firebase/functions/lib/events/live_meetings/breakouts/check_advance_meeting_guide.dart
mikewillems: You're right I was unclear here. But {diffusionStatement} is THE reason we compare on resolved rather than raw. Resolution substitutes the diffusion statement into an item, or drops the item entirely when there's no statement (i.e. prod), so resolved can be shorter than raw, so the last agenda item shown may not be the last raw item. The last item check has to use the same list the advance navigation uses (resolved) so the client (resolved) doesn't treat the last visible item as final and hide the ring while the backend (raw) still uses a timer for it, which would hide a countdown while it was happening. That mismatch is what I saw when testing and why this was a thing.

docs/pages/dev.md
mikewillems: 🙏

client/lib/features/events/features/live_meeting/features/meeting_guide/presentation/meeting_guide_card_presenter.dart
```diff
@@ -203,13 +203,25 @@ class MeetingGuideCardPresenter {
         _meetingGuideCardStore.isHoldingPendingAdvanceTransition;
   }
 
+  /// Whether [currentAgendaItemId] is the final agenda item. The backend finishes
+  /// the meeting immediately for the last item (no scheduled advance), so the
+  /// "moving onto the next agenda item" countdown must not be shown for it.
+  bool isLastAgendaItem(String? currentAgendaItemId) {
+    if (currentAgendaItemId == null) return false;
+    final items = _agendaProvider.resolvedAgendaItems;
+    return items.isNotEmpty && items.last.id == currentAgendaItemId;
```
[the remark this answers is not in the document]
johnnycrich: This is not correct.

client/lib/features/events/features/live_meeting/presentation/live_meeting_mobile_presenter.dart
```diff
@@ -207,13 +207,25 @@ class LiveMeetingMobilePresenter {
         _meetingGuideCardStore.isHoldingPendingAdvanceTransition;
   }
 
+  /// Whether [currentAgendaItemId] is the final agenda item. The backend finishes
+  /// the meeting immediately for the last item (no scheduled advance), so the
+  /// "moving onto the next agenda item" countdown must not be shown for it.
+  bool isLastAgendaItem(String? currentAgendaItemId) {
+    if (currentAgendaItemId == null) return false;
+    final items = _agendaProvider.resolvedAgendaItems;
+    return items.isNotEmpty && items.last.id == currentAgendaItemId;
```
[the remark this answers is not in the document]
johnnycrich: Again, I don't really understand this. The backend is definitely now using the resolved agenda items as a decider. Copilot is, as usual, being overly wordy, though, so maybe I'm misunderstanding this feedback.

johnnycrich (approved):
Nice work! Some good catches here. Also verifying that it indeed does solve some issues that I saw on staging.

PR #636, opened 2026-09-22
Fix Firefox DNS errors

mikewillems:
## What is in this PR?

Some participants (reported on Firefox/Safari) hit a generic error dialog when clicking "I'm ready" to advance a breakout agenda and when rejoining the main room after breakouts. Both actions call Firebase callable Cloud Functions. On the affected machines the cross-origin functions host (`us-central1-<project>.cloudfunctions.net`) fails DNS resolution (VPN, DNS filter, iCloud Private Relay, or a privacy/ad-blocking extension), so the callable client returns code `internal` / `INTERNAL`, which the app rendered as "Something went wrong". Firestore and Agora use different, reachable hosts, so the rest of the meeting looked fine.

## Changes in the codebase

1. `fwd callable failures (dns blocked) with useful message, drop prints`
   - `sanitizeError` now maps `internal` / `unavailable` / `Failed to fetch` /
     DNS-hostname errors to a new `networkRequestBlocked` message; removed the
     `Callable origin/uri` debug prints.
2. `retry transient network failures for GCF callables`
   - `callFunction` retries (2x, exponential backoff) only transient codes
     (`internal`, `unavailable`, `deadline-exceeded`) and raw network errors.
3. `translate networkRequestBlocked for es, zh, zh_Hant_TW`
   - Adds the new string to the three non-English arb files.
4. `route GCF callable functions from app origin`
   - Web callables in `sameOriginCallables` are called on the app origin via
     `/api/<name>` firebase.json rewrites, avoiding the cross-origin host and
     CORS preflight.
5. `Tests for callable network-failure notice, retry, and origin`
   - `sanitizeError` branch, retry predicate, and firebase.json/allowlist parity.
6. Added documentation / instructions to add callables DNS rewrite per-function.

None.

## Testing this PR

This bug was environment-dependent (client DNS/blocker), so I verified by blocking`*.cloudfunctions.net` manually w/ host list in Firefox to confirm that the "I'm ready" and rejoin cases: a. show the actionable message pre-fix and b. succeed after same-origin routing. They did locally.

Now I need us to test this from the prior affected clients (@mwmarino 😃 )

## PR Checklist

- [x] I have built locally and committed any updated freezed files.
- [x] For new Cloud Functions, I have added the function to `function-mapping.json`.
- [x] For new client-callable Cloud Functions, I have added an `/api/<FunctionName>` rewrite in `firebase.json` and the function to the `sameOriginCallables` allowlist.
- [x] I have ensured that all touched files have been formatted and are unmarked by the linter.
- [x] I have updated any relevant documentation for changed functionality.

## Additional information

- Same-origin routing uses a collision-safe `/api/` prefix so the rewrites
  cannot shadow any Flutter app route (e.g. `/joinEvent`).
- The client uses an allowlist (`sameOriginCallables`) rather than routing every
  callable. A callable missing a rewrite falls back to today's cross-origin
  behavior (no regression) instead of hitting the `ServeIndex` catch-all and
  breaking for everyone. `same_origin_callables_test` asserts the allowlist and
  the `/api/*` rewrites stay in exact sync.
- The parity test is VM-only (`@TestOn('vm')`) because it reads firebase.json
  from disk; the other two unit tests run on the chrome platform used by CI.

"Sentry Seer" pointed out bug locations. Reviewing with co-pilot.


client/lib/core/data/services/cloud_functions.dart
```diff
@@ -1,40 +1,79 @@
...
+
+    for (var attemptNum = 0;; attemptNum++) {
+      try {
+        return await attempt();
+      } catch (e) {
+        if (attemptNum >= _maxRetries || !isRetryableError(e)) {
+          rethrow;
+        }
+        await Future<void>.delayed(
+          Duration(milliseconds: 300 * (1 << attemptNum)),
```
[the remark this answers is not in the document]
mikewillems: This is way overblown - most of the reason this could happen is the DNS stuff we just fixed. Still, there's some useful guarding to do around this. Addressed in `1ec416`

client/lib/core/data/services/same_origin_callables.dart
```diff
@@ -0,0 +1,49 @@
...
+  'eventEnded',
+  'getCommunityCalendarLink',
+  'getCommunityCapabilities',
+  'getStripeBillingPortalLink',
+  'getStripeConnectedAccountLink',
+  'joinEvent',
+  'resolveJoinRequest',
+  'sendAnnouncement',
+  'sendEventMessage',
+  'toggleLikeDislikeOnMeetingUserSuggestion',
```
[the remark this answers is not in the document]
mikewillems: This is correct but pretty tiny / insignificant. Good hygiene and nice catch though. This was some of the OG Frankly code! Addressed in `5fb3f0`

client/lib/core/utils/error_utils.dart
```diff
@@ -43,7 +69,7 @@ String sanitizeError(String error) {
     return appLocalizationService.getLocalization().notAuthorized;
   }
   if (error.trim().toLowerCase() == 'INTERNAL'.toLowerCase()) {
-    return appLocalizationService.getLocalization().somethingWentWrong;
+    return appLocalizationService.getLocalization().networkRequestBlocked;
```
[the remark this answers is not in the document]
mikewillems: I should've caught this - good feedback. Addressed in `307b0e`.

client/test/lib/core/data/services/same_origin_callables_test.dart
```diff
@@ -0,0 +1,41 @@
+@TestOn('vm')
```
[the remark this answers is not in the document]
mikewillems: This is on point and the biggest real concern here. Added a drift test for `firebase.json` too. Addressed in `b1941a`

client/test/lib/core/data/services/same_origin_callables_usage_test.dart
```diff
@@ -0,0 +1,88 @@
...
+import 'package:client/core/data/services/same_origin_callables.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+// Guards against the case where the allowlist is ineffectual: a callable can be listed in
+// sameOriginCallables (and firebase.json) yet still call the cross-origin host
+// directly. Every allowlisted name must be invoked through CloudFunctions
+// .callFunction, and none may bypass it via a direct httpsCallable literal.
+// This catches any additional same-origin callables that were called as
+// external functions without adding them to the allowlist, which would break
+// the same-origin rewrite and repeat the Firefox-breaking erro.
```
johnnycrich, line 15: nit: typo `erro.`

client/test/lib/core/utils/error_utils_test.dart
```diff
@@ -0,0 +1,57 @@
...
+    genericMessage = l10n.somethingWentWrong;
+  });
+
+  tearDownAll(() async {
+    await GetIt.instance.reset();
+  });
+
+  void expectNetwork(String error) =>
+      expect(sanitizeError(error), networkMessage);
+  void expectGeneric(String error) =>
```
johnnycrich, line 28: Nice test organization!

.github/workflows/test_client.yaml
```diff
@@ -25,6 +25,9 @@ jobs:
           filters: |
             client:
               - 'client/**'
+              # The /api callable rewrites the VM parity test checks against the
```
johnnycrich, line 28: I find this a little unclear. The intent of including `firebase.json` in the filter is hard to understand. I would explicitly state that that file contains the /api callable rewrite config used by the parity/allowlist test, and that changes there should trigger the client test to detect drift.

client/lib/core/utils/error_utils.dart
```diff
@@ -19,7 +19,32 @@ String? _firebaseAuthCodeMessage(String code) {
   }
 }
 
+/// True for errors that never reached the server (DNS/VPN/relay/ad-block).
+bool _looksLikeNetworkFailure(String error) {
+  final lower = error.toLowerCase();
+  const needles = [
```
johnnycrich, line 25: This made me LOL, but feels a little non-specific. Maybe `networkFailureIndicators`?

docs/pages/hosting.md
```diff
@@ -115,6 +115,38 @@ This is configured by the catch-all rewrite in `firebase.json`:
 }
 ```
 
+### Same-origin callable routing (`/api/*`)
```
johnnycrich, line 118: Great work with this documentation.

johnnycrich (changes requested):
Tested with advised URL blocking and worked for me (rewrite succeeds), all other systems seem nominal. Just a couple things to look at!

client/test/lib/core/data/services/same_origin_callables_usage_test.dart
mikewillems: Good catch! For documentation-only changes, they don't need re-review after changing, so best to note and approve (the author can be trusted to make the changes).

client/lib/core/utils/error_utils.dart
mikewillems: "looksLike" is actually descriptive here. Could use "presentsAs", but that's more wordy and less direct. Quite specifically, this means that it has that signature.

docs/pages/hosting.md
mikewillems: Thanks! I try :P

johnnycrich (approved):

